PR #7, opened 2017-04-18
Performance improvement sinkhorn

aje:
Doing the computation this way is equivalent and allows to reduce the space complexity required from O(max(a, b)^2) to O(a*b) (especially usefull to transport a small number of sources example to a lot of target)

This also allows to decrease the computation time.

### DIFF
--- a/ot/bregman.py
+++ b/ot/bregman.py
@@ -108,7 +108,8 @@ def sinkhorn(a,b, M, reg, numItermax = 1000, stopThr=1e-9, verbose=False, log=Fa
     K = np.exp(-M/reg)
     #print(np.min(K))
 
-    Kp = np.dot(np.diag(1/a),K)
+    Kp = (1/a).reshape(-1, 1) * K
+
     transp = K
     cpt = 0
     err=1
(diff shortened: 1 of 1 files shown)


aje:
I have only github home.
My GPU implementation of the Sinkhorn algo is based on another library (cudamat https://github.com/cudamat/cudamat) to perform the long matrix computations on the GPU (multiplication, pairwise euclidean distance, ...). I however had to modify this library to support 64 float point numbers; I will try to see if I can make a self contained GPU version before submitting it.

AjayTalati:
Hi @aje, my friend coded up Sinkhorn in PyTorch, maybe that could be helpful to you?

https://github.com/t-vi/pytorch-tvmisc/blob/master/wasserstein-distance/Pytorch_Wasserstein.ipynb

We are both interested in application of optimal transport in deep learning - it's nice to see others are working on similar things too :+1:

rflamary:
OK great,

GPU implementation is of strong interest to us. it can be added as a sub-module with its own dependencies without problems (for instance ot.gpu for cudamat and/or ot.Pytorch for the Pytorch implementations). The only requirement would be compatible functions calls which is the case in the notebook above. 

Do you have some ideas/numbers for the speedup. Should be important in float32 and average in float64 from my experience.

AjayTalati:
OK, I'll try my best,

It would be a lot of fun if there was some sort of real data test problem to use it on? I've seen applications to the "word movers distance",

https://github.com/mkusner/wmd

in previous publications by the contributors of the POT library. If you have any code from such papers, that would be really helpful - things are easier to understand and more motivating when there's an intuitive application :+1: 

It would be nice to test the code verses, previously published results?

Thanks for your help!

PR #10, opened 2017-04-21
performance improvement sinkhorn lpl1

aje:
- instead of updating individually for each target examples, update for all target examples at once using numpy functions. This allows for a faster computation (for me, divided by 4 on 3000*100 random matricies and random labels in [0,1]).

 - if I understoud correctly, a value of -1 in the array labels_a meant that we didn't have a label for this example. But in machine learning, we often encounter the binary case where we say we have the positive class (+1) and negative class (-1); thus with a dataset like this, the algorithm wouldn't work as expected. I replaced the default value for 'no label' to '-99' instead of '-1', and I added a parameter to modify it.

### DIFF
--- a/ot/da.py
+++ b/ot/da.py
@@ -11,11 +11,7 @@ from .optim import cg
 from .optim import gcg
 
 
-def indices(a, func):
-    return [i for (i, val) in enumerate(a) if func(val)]
-
-
-def sinkhorn_lpl1_mm(a,labels_a, b, M, reg, eta=0.1,numItermax = 10,numInnerItermax = 200,stopInnerThr=1e-9,verbose=False,log=False):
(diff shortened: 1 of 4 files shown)


ncourty:
Dear Aje,
Many thanks for the pull request. As a matter of fact, you have highlighted 
a problem in this function. The handling of unlabelled data is not possible in this method, and what was implemented was not correct. Setting a majoration to zero in the current code will not have the correct behaviour as it will tend to favorize in the transport plan connections between unlabelled data. This method should only be used with fully labelled distributions. Before I do the merge,
can you please update your code to remove the case where unlabelled data are taken into account ? 
Thanks,

aje:
Ok I will change this. this will make the code easier to understand 👍

PR #13, opened 2017-07-11
[MRG] flake8 + pimp example figures

agramfort:
@rflamary @ncourty any opinion on this? shall I continue?

big questions are
- ok to adopt pep8?
- do you agree to use py.test for more standard testing mechanism?

don't merge yet. Just tell me what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# Mac stuff
+.DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ before_install:
 # command to install dependencies
 install:
   - pip install -r requirements.txt
+  - pip install flake8 pytest
   - python setup.py install
-# command to run tests
-script: python test/test_load_module.py -v
+# command to run tests + check syntax style
+script:
+  - python test/test_load_module.py -v
+  - flake8 .
+  - py.test ot test
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ install:
 # command to run tests + check syntax style
 script:
   - python test/test_load_module.py -v
-  - flake8 .
+  - flake8 examples/ ot/ test/
   - py.test ot test
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ install:
 script:
   - python test/test_load_module.py -v
   - flake8 examples/ ot/ test/
-  - py.test ot test
+  # - py.test ot test
--- a/examples/plot_OTDA_2D.py
+++ b/examples/plot_OTDA_2D.py
@@ -11,110 +11,112 @@ import matplotlib.pylab as pl
 import ot
 
 
-
 #%% parameters
 
-n=150 # nb bins
+n = 150  # nb bins
 
-xs,ys=ot.datasets.get_data_classif('3gauss',n)
-xt,yt=ot.datasets.get_data_classif('3gauss2',n)
+xs, ys = ot.datasets.get_data_classif('3gauss', n)
+xt, yt = ot.datasets.get_data_classif('3gauss2', n)
 
-a,b = ot.unif(n),ot.unif(n)
+a, b = ot.unif(n), ot.unif(n)
 # loss matrix
-M=ot.dist(xs,xt)
-#M/=M.max()
+M = ot.dist(xs, xt)
+# M/=M.max()
 
 #%% plot samples
 
 pl.figure(1)
-
-pl.subplot(2,2,1)
-pl.scatter(xs[:,0],xs[:,1],c=ys,marker='+',label='Source samples')
+pl.subplot(2, 2, 1)
+pl.scatter(xs[:, 0], xs[:, 1], c=ys, marker='+', label='Source samples')
 pl.legend(loc=0)
 pl.title('Source  distributions')
 
-pl.subplot(2,2,2)
-pl.scatter(xt[:,0],xt[:,1],c=yt,marker='o',label='Target samples')
+pl.subplot(2, 2, 2)
+pl.scatter(xt[:, 0], xt[:, 1], c=yt, marker='o', label='Target samples')
 pl.legend(loc=0)
 pl.title('target  distributions')
 
 pl.figure(2)
-pl.imshow(M,interpolation='nearest')
+pl.imshow(M, interpolation='nearest')
 pl.title('Cost matrix M')
 
 
 #%% OT estimation
 
 # EMD
-G0=ot.emd(a,b,M)
+G0 = ot.emd(a, b, M)
 
 # sinkhorn
-lambd=1e-1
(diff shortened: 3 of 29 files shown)


rflamary:
Hello @agramfort ,

- I agree with pep8 adoption even though some of the rules are dated (80 characters is tough when you have classes). In fact I was working with autopep8 to clean this with no headaches but I think you are more competent than me here.

- py.test  is also better and we will add some proper tests at least for the main functions quickly. Clearly those with GPU cannot build and run on travis so they should be optional (i dont kow how to do that with pytest other than removing the files).

Also I know I use pl instead of plt but it felt so much better at the time (still feels better trust me). Still i understand the need to make the examples more friendly to newcomers.

@ncourty  how so you feel about all that?

ncourty:
Hello @agramfort and @rflamary,

- Clearly we should go for a strict application of pep8 recommendations, even if I do agree with @rflamary that some of them are outdated (I've read several passionate discussions on the 80 chars here and there). Being compliant with pep8 standards, even if too restrictive, is a big plus IMO

- Regarding pytest, I also do agree that it is one of the best option available for unitary tests and continuous integration, so I think it's also good practice. We need to level up on how to use it though, but it does not seem too complicated.

- Regarding pl. vs plt., I have seen both versions on many codes. @agramfort, is there a standard here (maybe part of pep8 ?)

agramfort:
hey,

so I made progress. I pushed full flake8 valid code.

I'll need you to refactor the tests to they are functions that start with
test_

then we can call SkipTest if cudamat is not installed.

Some notes from my reading:

- why not using cg from scipy.optimize?
- sklearn calls make_ functions that make simulated data. So I propose to
rename get_ -> make_ for datasets
- add references to papers in examples (it's not always there)
- don't put params in fit in the estimators. Put parameters in the init
- don't mix CamelCase and snake_case eg OTDA_mapping_kernel ->
OTDAMappingKernel etc.
- why gpuarray and not cupy which seem more mature? It should also allow
you to have the same code for CPU and GPU code.

feedback welcome.

A

rflamary:
Hello,

Thank you for the work, it looks great I just have small remarks:
- full flake8 valid is awesome !
- you did no change all pl to plt, probably because we did no concrge with @ncourty. Sp we need to do it quickly I suppose.

Regarding your remarks:
- CG here stands for conditional gradient and not conjugate gradient as implemented in scipy.
- OK for switch to make but all examples and notebooks have to be updated also.
- Example do not have a proper description yet and I agree we need to complete it with
- I always thought this choice (params in init) in sklearn was weird since it make it more complicated  to perform warm start with a slightly different regularization (in a reg path for instance) value which is common practice in ML. I'm sure you have good reasons for it though. my problem here is that it breaks the code for people who already use the toolbox and I'd like to keep the young and still small community.. Maybe do both for a while with a Deprecated Warning?
-  I agree we should stick to Camel or snake but i personally prefer snake and most of the toolbox is in snake, do you suggest the change for Classes only?, is it a common thing to do ?
- I agree that the choice for cudamat is probably not the best. We were planing with Nico to switch to a more future proof code with pycuda.

I'll work on the test next week when I don't have NIPS review to write.

agramfort:
> full flake8 valid is awesome !

:)

> you did no change all pl to plt, probably because we did no concrge with
@ncourty. Sp we need to do it quickly I suppose.

yes I used pl so you're still in your confort zone :)

> Regarding your remarks:
>
> CG here stands for conditional gradient and not conjugate gradient as
implemented in scipy.

arfff good point. What it tells you is that cg is probably not explicit
enough. Maybe conditional_gradient is not that long to type :)

> OK for switch to make but all examples and notebooks have to be updated
also.

arfff. But I thought you were using sphinx-gallery. WIth sphinx-gallery you
should not ship notebooks. They are produced when you build the doc.

> Example do not have a proper description yet and I agree we need to
complete it with
> I always thought this choice (params in init) in sklearn was weird since
it make it more complicated to perform warm start with a slightly different
regularization (in a reg path for instance) value which is common practice
in ML.

for this we use the warm_start parameter

clf = Classifier(..., warm_start=True)
clf.fit(X, y)
clf.set_param(C=C_new)  # or just clf.C = C_new
clf.fit(X, y)


> I'm sure you have good reasons for it though. my problem here is that it
breaks the code for people who already use the toolbox and I'd like to keep
the young and still small community.. Maybe do both for a while with a
Deprecated Warning?

yes you need deprecation cycles. But the earlier you do it the better. If
it's too painful just deprecate a full class and create a new one with a
better name.

> I agree we should stick to Camel or snake but i personally prefer snake
and most of the toolbox is in snake, do you suggest the change for Classes
only?, is it a common thing to do ?

it's not a matter of preference but conventions :)

yes you should use CamelCase for classes and snake_case for functions and
methods.

> I agree that the choice for cudamat is probably not the best. We were
planing with Nico to switch to a more future proof code with pycuda.

pycuda is old and not that maintained anymore. cupy is much more used these
days and more actively maintained.

> I'll work on the test next week when I don't have NIPS review to write.

:)

rflamary:
Hello Alexandre,

We met with Nico and decided on a few things discussed in the following

2017-07-13 14:44 GMT+02:00 Alexandre Gramfort <notifications@github.com>:
>
>
> > you did no change all pl to plt, probably because we did no concrge with
> @ncourty. Sp we need to do it quickly I suppose.
>
> yes I used pl so you're still in your confort zone :)
>

OK so we decided to stick to pl since we import pylab and not pyplot and
all
the modules in pot are two letters.



>
> > Regarding your remarks:
> >
> > CG here stands for conditional gradient and not conjugate gradient as
> implemented in scipy.
>
> arfff good point. What it tells you is that cg is probably not explicit
> enough. Maybe conditional_gradient is not that long to type :)
>

ok for renaming the function and also for generalized cg.  temporary we
will keep cg=conditional_gradient in order to keep the compatibility and
will put deprecation warning in the future.


>
> > OK for switch to make but all examples and notebooks have to be updated
> also.
>
> arfff. But I thought you were using sphinx-gallery. WIth sphinx-gallery you
> should not ship notebooks. They are produced when you build the doc.
>

OK but until recently I did not know how to do proper titles and text in
the python code so that you have a notebook with cells instead of one giant
cell as in most sklearn examples. This will take some time but we will
switch to python only examples as soon as we transcript the text and titles
from the notebooks.



Params in init:
OK for renaming  the classes in CamelCase aznd moving the parameters to the
init phase. We will also add a set_param function in order to be more
compatible with sklearn


>
> > I agree that the choice for cudamat is probably not the best. We were
> planing with Nico to switch to a more future proof code with pycuda.
>
> pycuda is old and not that maintained anymore. cupy is much more used these
> days and more actively maintained.
>

yeah ok for cupy we will look into that

agramfort:
ok fine with me. So how do you want to proceed with this PR?
I rebased and revert the plt

did you look at the diff?

rflamary:
Yeah we are OK with the merge but just wait for travis to merge.

Seems from early results that some PEP8 error remain. 

Thank you again

agramfort:
all green !

PR #14, opened 2017-07-19
include LICENSE in MANIFEST.in

dougalsutherland:
It's generally good practice to distribute the text of the license with your package; this change makes sure that your `LICENSE` file ends up in the pypi tar files. (It currently doesn't.)

(This came up because I'm putting together a [conda-forge](http://conda-forge.github.io) recipe for POT, which ideally wants a file with the license text.)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 graft ot/lp/
 include README.md
+include LICENSE
 include ot/lp/core.h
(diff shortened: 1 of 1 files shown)


rflamary:
Hello Dougal

Thank you for this pull request and for taking care of the conda-forge for POT. 

I accept the merge.

Rémi

Issue #21, opened 2017-07-26
Sklearn compliant datasets functions


rflamary:
We should rename the datasets.get_* function to datasets.get_* in order to be more sklearn compliant.

Also it should be possible to give the rng as input as in sklearn.

rflamary:
I'm closing this issue since it has been resolved when merging PR #49

PR #22, opened 2017-07-28
Domain adaptation Classes

Slasnista:
- first proposal of DA class structure
- BaseEstimator: OTDA wrapper (does not work as a stand-alone but implements the methods common to any OTDA algorithm)
- SinkhornTransport: implements Sinkhorn algorithm for OTDA
- try doc strings compliant with numpy requirements


rflamary:
Thank you @Slasnista , I will have a look this week end or monday.

Maybe creating a compliant BaseEstimator Class as @agramfort  proposed is a nice way to avaoid sklearn dependency.

rflamary:
Hello, I left a few comments in response to @ngayraud.  

@Slasnista I think this is a good start but we need to agree on a typical use case. below how I see the class working:

```python

tr=EMDTransport(out_of_sample_map='ferradans',weight_est='uniform')

tf.fit(Xs,Xt) # learn G matrix and store it, also store data

Xs_trans=tr.transform(Xs) # barycentric mapping
Xs_trans=tr.transform() # barycentric mapping with stored data
Xs2_trans=tr.transform(Xs2) # barycentric mapping + Ferradans since out of sample

```

Slasnista:
@rflamary thanks for the comments. I took them into account and I will update my pr.

For the use cases I agree with you on most of them. Yet even it's a bit redundant, it looks weird to not pass data to the transform method.

So if we opt for this solution, we need to write a detailed example to explain the differences between the different version of transform. I will look at your examples to update them accordingly to the classes structure.

rflamary:
@Slasnista OK, as a second though I agree with you it's weird with no params. Let's not do it.

Still I think it is very important to handle automatically the known Xs and new our of samples Xs2.

Rémi

Slasnista:
**changes**
- as @agramfort suggested, I removed dependency on sklearn's BaseEstimator
- removed also dependency on sklearn's pairwise_distance as @ngayraud proposed
- added in `__init__`a parameter `density_estimation` which takes a function as input. (default: uniform)
- stored Xs and Xt used for coupling estimation as attributes.
- transform and inverse_transform modified: they check if their respective input has already been seen: if yes applies barycentric mapping, if no they will apply out of samples mapping (not implemented yet)

ot/da.py
```diff
@@ -921,21 +921,153 @@ def predict(self, x):
...
 
-from sklearn.base import BaseEstimator
-from sklearn.metrics import pairwise_distances
+# from sklearn.base import BaseEstimator
+# from sklearn.metrics import pairwise_distances
+
+##############################################################################
+# adapted from scikit-learn
+
+import warnings
```
rflamary: feel free to put it at the top of the file

ot/da.py
```diff
@@ -921,21 +921,153 @@ def predict(self, x):
...
+        class_name = self.__class__.__name__
+        return '%s(%s)' % (class_name, _pprint(self.get_params(deep=False),
+                                               offset=len(class_name),),)
+
+    # __getstate__ and __setstate__ are omitted because they only contain
+    # conditionals that are not satisfied by our objects (e.g.,
+    # ``if type(self).__module__.startswith('sklearn.')``.
+
+
+def distribution_estimation_uniform(X):
```
rflamary: note that there exist the function ot.utils.unif that does exactly the same thing

maybe a simple
```python
from .utils import unif
distribution_estimation_uniform=unif
```
would do the trick

ot/da.py
```diff
@@ -960,18 +1092,19 @@ def fit(self, Xs=None, ys=None, Xt=None, yt=None):
...
-            mu_s = np.ones(Xs.shape[0]) / float(Xs.shape[0])
-            mu_t = np.ones(Xt.shape[0]) / float(Xt.shape[0])
-        else:
-            print("TODO: implement kernelized approach")
+        mu_s = self.distribution_estimation(Xs)
+        mu_t = self.distribution_estimation(Xt)
+
+        # store arrays of samples
+        self.Xs = Xs
+        self.Xt = Xt
```
rflamary: Remove sinkhorn from the BAseEstimator and put it in the SinkhornTransport please.

Keep all teh rest so that we can do super as still have  all the storing and density estimation.

ot/da.py
rflamary: a no it does not wrok but you still can use unif at the return line

ot/da.py
```diff
@@ -1114,7 +1255,10 @@ class SinkhornTransport(BaseTransport):
 
     def __init__(self, reg_e=1., mode="unsupervised", max_iter=1000,
                  tol=10e-9, verbose=False, log=False, mapping="barycentric",
-                 metric="sqeuclidean", distribution="uniform"):
+                 metric="sqeuclidean",
+                 distribution_estimation=distribution_estimation_uniform,
```
ngayraud: Ok good. It is much simpler and allows the user to move around freely with the distribution estimation.

ot/da.py
```diff
@@ -1123,8 +1267,9 @@ def __init__(self, reg_e=1., mode="unsupervised", max_iter=1000,
         self.log = log
         self.mapping = mapping
         self.metric = metric
-        self.distribution = distribution
+        self.distribution_estimation = distribution_estimation
         self.method = "sinkhorn"
+        self.out_of_sample_map = out_of_sample_map
 
     def fit(self, Xs=None, ys=None, Xt=None, yt=None):
         """Build a coupling matrix from source and target sets of samples
```
ngayraud: I see that in every fit and transform, all of the parameters are initialized to None. However, I think that at least Xs (and Xt for the inverse transform) should be required.

Slasnista:
Hi,
- I added a test function for SinkhornTransport()
- I slightly changed SinkhornTransport() class. There are still some changes needed but we are converging.

Here some questions:
- could we move the test folder in the ot foler ?
- why are there so many things in the `__init__` of ot folder ?
- could we change the names of parameters and variables in da functions ? (I fear I, G, M... might be tricky for debugging in few months ?)
- do you have a way to check the rendering of doc string ? I'd like to make new class docstrings compliant with your guidelines now, so that we won't have to spend extra sessions documenting code ;)

ngayraud:
Here is a bit of code you can add for a simple EMD transport. I hope I have it right.
Also from what I remember, if we set log = true, the methods return the log as well. So we need an "If log statement (see code below)




    class EMDTransport(BaseTransport):
        """Domain Adapatation OT method based on Earth Mover's Distance
        Parameters
        ----------
        mode : string, optional (default="unsupervised")
            The DA mode. If "unsupervised" no target labels are taken into account
            to modify the cost matrix. If "semisupervised" the target labels
            are taken into account to set coefficients of the pairwise distance
            matrix to 0 for row and columns indices that correspond to source and
            target samples which share the same labels.
        mapping : string, optional (default="barycentric")
            The kind of mapping to apply to transport samples from a domain into
            another one.
            if "barycentric" only the samples used to estimate the coupling can
            be transported from a domain to another one.
        metric : string, optional (default="sqeuclidean")
            The ground metric for the Wasserstein problem
        distribution : string, optional (default="uniform")
            The kind of distribution estimation to employ
        verbose : int, optional (default=0)
            Controls the verbosity of the optimization algorithm
        log : int, optional (default=0)
            Controls the logs of the optimization algorithm
        Attributes
        ----------
        References
        ----------
        .. [1] N. Courty; R. Flamary; D. Tuia; A. Rakotomamonjy,
               "Optimal Transport for Domain Adaptation," in IEEE Transactions
               on Pattern Analysis and Machine Intelligence , vol.PP, no.99, pp.1-1
        """

        def __init__(self, mode="unsupervised", verbose=False, 
                     log=False, metric="sqeuclidean",
                     distribution_estimation=distribution_estimation_uniform,
                     out_of_sample_map='ferradans'):

            self.mode = mode
            self.verbose = verbose
            self.log = log
            self.metric = metric
            self.distribution_estimation = distribution_estimation
            self.out_of_sample_map = out_of_sample_map

        def fit(self, Xs, ys=None, Xt=None, yt=None):
            """Build a coupling matrix from source and target sets of samples
            (Xs, ys) and (Xt, yt)
            Parameters
            ----------
            Xs : array-like of shape = [n_source_samples, n_features]
                The training input samples.
            ys : array-like, shape = [n_source_samples]
                The class labels
            Xt : array-like of shape = [n_target_samples, n_features]
                The training input samples.
            yt : array-like, shape = [n_labeled_target_samples]
                The class labels
            Returns
            -------
            self : object
                Returns self.
            """

            self = super(EMDTransport, self).fit(Xs, ys, Xt, yt)

            # coupling estimation
            if self.log:
                        self.gamma_, self.log_ = emd(
                            a=self.mu_s, b=self.mu_t, M=self.Cost, 
                            verbose=self.verbose, log=self.log)
            else:
                        self.gamma_= emd(
                            a=self.mu_s, b=self.mu_t, M=self.Cost, 
                            verbose=self.verbose, log=self.log)

test/test_da.py
```diff
@@ -6,6 +6,57 @@
...
 import numpy as np
 import ot
+from numpy.testing.utils import assert_allclose, assert_equal
+from ot.datasets import get_data_classif
+from ot.utils import unif
+
+np.random.seed(42)
+
+
+def test_sinkhorn_transport():
```
Slasnista: thank you for the reply, it's good for me. I will modify the name of the test function

Slasnista:
@ngayraud thank you for the code ! I adapt it and put it inside the pr ;)

just for if we can solve it like it 

    # coupling estimation
    returned_ = emd(
        a=self.mu_s, b=self.mu_t, M=self.Cost, 
        verbose=self.verbose, log=self.log)
    if self.log:
         self.coupling_, self.log = returned_
    else:
        self.coupling_ = returned_

That does not change much, but we do not have the coupling estimation in a loop any more.

Btw, do all the other methods return log in addition to couplings ?

rflamary:
Emd do not accept log or verbose. It's juste à frapper for c code.



Le 1 août 2017 13:01, "Stanislas Chambon" <notifications@github.com> a
écrit :

Slasnista:
@rflamary thx, the test gave me this error ;) I removed the log and verbose input to emd for the moment.

ngayraud:
I think that all the sinkhorn based transports return a log! I was unsure about the emd one though.

Slasnista:
Hi,
I did some update to the PR:
- add a test for fit_transform method and correction of a bug for this method (did not return self)
- add 2 new classes with class regularization `SinkhornLpl1Transport()` and `SinkhornL1l2Transport()`
- added few in lines in .fit() method of BaseTransport() class to support semi supervised domain adaptation. It implies that the labeled sources (resp. target) samples are in the first rows (resp. columns) + added a test to check that when semi supervised is selected but no labels (neither ys, yt, nor both) unsupervised DA is performed instead of semi-supervised

What do you think about these new modifications ?

rflamary:
Hello @Slasnista,

Those are nice new features.

I just have a few questions/todos:

- Why is the method necessary? I think the fact that the ys and yt are set to None is enough to know if semi supervised should be used. No need to have two signal that can contradict...
- Also what is that about the supervised sample being the first ones? Couldn't we use ys/yt to define the supervised ones (0..N) and usnupervised (-1) ?
- Semi supervised as in the PAMI paper does not work with O cost when the class is the same. It uses infinite (or very large) cost when the label is different. This is a versy different approach based on DO-NOT-LINK constratints and still provide OT inside classes which is no done when the loss in 0.

- Finally I think the last piece before merging is coding a proper out of sample as done here:
https://github.com/Slasnista/POT/blob/0b005906f9d78adbf4d52d2ea9610eb3fde96a7c/ot/da.py#L712
Note that it's just a few lines but one should perform it on minibach in order to limit the memory requirement and hence it should Have a loop.

- I know I am never happy but the interpolation (linear and kernel regression methods are also missing). 

- Once all the classes are done properly, feel free to add a warning to OTDA class that tell about the future deprecation and the new classes. I will add both of you to the contributor list BTW.

This will be a very nice merge thank you again for all the work .

Slasnista:
@rflamary:
1. ok for me, this way it will not be too redundant.
2. ok, the way it's written works too independently of the position of the labeled samples so nothing to change
3. ok, but how do you set this infinite value ? 10**2, 10**3, something based on the max value of the cost matrix ? Do you have a predefined rule for doing this ?

I address the following points in a a second time ;)

rflamary:
@Slasnista nice to see we agree. Point 3 is the tough one ;).

For sinkhorn and group lasso regularization we set it to infinity (np.inf) and under constraint qualification conditions (there exist a solution that is not infinite), the algorithm converges to a solution (with zerso mass on the do-not-link links).

For emd solver (and conjugate gradient that uses emd) I don't think the solver is going to appreciate infinity so I would use as you propose the max value of the cost matrix plus something maybe epsilon(max(Cost)). I think it should be a parameter anyway with the max/inf taken if set to None.

Take your time for the remaining I will be in vacation next week and probably won't fine time to do a proper final review anyway.

Slasnista:
Thank you for the fast answer. Ok, good so if I sum it up:

When you modify your cost matrix, you both:
- set the coefficient corresponding to source and target samples with same labels to 0
- set the coefficient corresponding to source and target samples with different labels to np.infty or something big depending on the Class.

Is that right ?

No worries, I would be off the next coming 2 weeks but I would like to clean some stuffs before going on holiday ;-)

ot/da.py
```diff
@@ -843,10 +909,672 @@ def predict(self, x):
...
+    """
+
+    def __init__(self, mode="unsupervised", verbose=False,
+                 log=False, metric="sqeuclidean",
+                 distribution_estimation=distribution_estimation_uniform,
+                 out_of_sample_map='ferradans'):
+
+        self.mode = mode
+        self.verbose = verbose
+        self.log = log
```
ngayraud: Since there is no log, or verbosity level, we can remove it from here.

rflamary:
@Slasnista actually no you just use the do not link. When you set the transport cost to zero you actually make the problem badly conditioned since every sample from a given class can be assigned to every sample from the same class for the same cost.

You still want to do OT (not permute samples for instance) and for that you need to keep the loss. Maybe using zero could lead to nice stuff but that is not what was proposed in the PAMI paper.

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+                                   "specify their parameters in the signature"
+                                   " of their __init__ (no varargs)."
+                                   " %s with constructor %s doesn't "
+                                   " follow this convention."
+                                   % (cls, init_signature))
+        # Extract and sort argument names excluding 'self'
+        return sorted([p.name for p in parameters])
+
+    def get_params(self, deep=True):
+        """Get parameters for this estimator.
```
agramfort: insert empty line

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        Parameters
+        ----------
+        deep : boolean, optional
+            If True, will return the parameters for this estimator and
+            contained subobjects that are estimators.
```
agramfort: insert empty line

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+
+            # XXX: should we rather test if instance of estimator?
+            if deep and hasattr(value, 'get_params'):
+                deep_items = value.get_params().items()
+                out.update((key + '__' + k, val) for k, val in deep_items)
+            out[key] = value
+        return out
+
+    def set_params(self, **params):
+        """Set the parameters of this estimator.
```
agramfort: insert empty line

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        The method works on simple estimators as well as on nested objects
+        (such as pipelines). The latter have parameters of the form
+        ``<component>__<parameter>`` so that it's possible to update each
+        component of a nested object.
```
agramfort: insert empty line

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+    # ``if type(self).__module__.startswith('sklearn.')``.
+
+
+def distribution_estimation_uniform(X):
+    """estimates a uniform distribution from an array of samples X
+
+    Parameters
+    ----------
+    X : array-like of shape = (n_samples, n_features)
+        The array of samples
```
agramfort: insert empty line

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+    # conditionals that are not satisfied by our objects (e.g.,
+    # ``if type(self).__module__.startswith('sklearn.')``.
+
+
+def distribution_estimation_uniform(X):
+    """estimates a uniform distribution from an array of samples X
+
+    Parameters
+    ----------
+    X : array-like of shape = (n_samples, n_features)
```
agramfort: X : array-like, shape (n_samples, n_features)

to be consistent with numpy

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        The array of samples
+    Returns
+    -------
+    mu : array-like, shape = (n_samples,)
```
agramfort: mu : array-like, shape (n_samples,)

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        The uniform distribution estimated from X
+    """
+
+    return unif(X.shape[0])
```
agramfort: should you really invent a function name for 1 line of numpy?

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        yt : array-like, shape = (n_labeled_target_samples,)
+            The class labels
+        Returns
+        -------
+        self : object
+            Returns self.
+        """
+
+        # pairwise distance
+        self.Cost = dist(Xs, Xt, metric=self.metric)
```
agramfort: any attribute changed in fit function should have trailing _
and caps should be avoided un var names. So I would write:

self.cost_ = dist(Xs, Xt, metric=self.metric)

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        Controls the verbosity of the optimization algorithm
+    log : int, optional (default=0)
+        Controls the logs of the optimization algorithm
+    limit_max: float, optional (default=10)
+        Controls the semi supervised mode. Transport between labeled source
+        and target samples of different classes will exhibit an infinite cost
+        (10 times the maximum value of the cost matrix)
+    Attributes
+    ----------
+    Coupling_ : the optimal coupling
```
agramfort: Coupling_ -> coupling_

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        Stop threshold on error (inner CG solver) (>0)
+    verbose : bool, optional (default=False)
+        Print information along iterations
+    log : bool, optional (default=False)
+        record log if True
+
+    Attributes
+    ----------
+    Coupling_ : the optimal coupling
+    Mapping_ : the mapping associated
```
agramfort: Mapping_ -> mapping_

ot/da.py
```diff
@@ -843,10 +909,807 @@ def predict(self, x):
...
+        self.inner_tol = inner_tol
+        self.log = log
+        self.verbose = verbose
+
+    def fit(self, Xs=None, ys=None, Xt=None, yt=None):
+        """Builds an optimal coupling and estimates the associated mapping
+        from source and target sets of samples (Xs, ys) and (Xt, yt)
+        Parameters
+        ----------
+        Xs : array-like of shape = (n_source_samples, n_features)
```
agramfort: = are not needed in shape descriptions

test/test_da.py
```diff
@@ -6,6 +6,263 @@
 
 import numpy as np
 import ot
+from numpy.testing.utils import assert_allclose, assert_equal
```
agramfort: import numpy stuff before ot

test/test_da.py
```diff
@@ -6,6 +6,263 @@
...
+from ot.datasets import get_data_classif
+from ot.utils import unif
+
+np.random.seed(42)
```
agramfort: you should not need global random seeds

test/test_da.py
```diff
@@ -6,6 +6,263 @@
...
+    clf.fit(Xs=Xs, ys=ys, Xt=Xt)
+    n_unsup = np.sum(clf.Cost)
+
+    # test semi supervised mode
+    clf = ot.da.SinkhornLpl1Transport()
+    clf.fit(Xs=Xs, ys=ys, Xt=Xt, yt=yt)
+    assert_equal(clf.Cost.shape, ((Xs.shape[0], Xt.shape[0])))
+    n_semisup = np.sum(clf.Cost)
+
+    assert n_unsup != n_semisup, "semisupervised mode not working"
```
agramfort: use assert_true from nose

test/test_da.py
```diff
@@ -6,6 +6,263 @@
...
+    clf.fit(Xs=Xs, ys=ys, Xt=Xt)
+    n_unsup = np.sum(clf.Cost)
+
+    # test semi supervised mode
+    clf = ot.da.SinkhornL1l2Transport()
+    clf.fit(Xs=Xs, ys=ys, Xt=Xt, yt=yt)
+    assert_equal(clf.Cost.shape, ((Xs.shape[0], Xt.shape[0])))
+    n_semisup = np.sum(clf.Cost)
+
+    assert n_unsup != n_semisup, "semisupervised mode not working"
```
agramfort: here

test/test_da.py
rflamary: Actually we use pytest that uses assert and maybe we should limit the dependencies.

agramfort:
Ok keep it as it is then

ot/da.py
ngayraud: I am going to guess that it it serves the puspose of allowing the user to provide his own function (kernel estimation for example).

Slasnista:
Hi,
Here a recap of what has been done and remains to do:

**DONE**:
- `BaseEstimator` class introduced accordingly to scikit-learn
- `BaseTransport` class (from `BaseEstimator`). It implements `.fit()`, `.transform()`, `.inverse_transform()`, `.fit_transform()`
    - `.transform()`, `.inverse_transform()` have out of sample transform accordingly to _Ferrandans et al. 2014_
- `EMDTransport`, `SinkhornTransport`, `SinkhornLpL1Transport`, `SinkhornL1l2Transport` classes (from `BaseTransport`)
- `MappingTransport` (from `BaseEstimator`). It supports linear and kernel mappings. It has methods `.fit()` and `.transform()`.

**TODO**:
- adapt out of samples transform to make it work by batch.
- test function for `MappingTransport`
- deprecation warning for old classes

If you see something else, don't hesitate to tell me ;-)

Slasnista:
I completed all the modifications we talked about.

I think we should grasp this occasion to think about the algorithms' outputs. In my opinion:
1. functions which compute couplings and or mappings should always return `logs_`.
2. functions which compute couplings and mappings should always return `coupling_, mapping_, bias_, logs_`. I would prefer return a zero vector instead of having an output of varying size depending of the bias.

Point **2.** is not a burning issue but **1.** is one.

Right now, instantiating an object with `logs=True` results in an error. I propose to remove logs from `__init__` and to let all `.fit` methods create an attribute `.logs_` which will contain the logs. What do you think about it ?

rflamary:
Hello @Slasnista,

Thank you for the new commits that handle all the things we discussed. I will do a proper code review but in the meantime you need to update the examples in the examples folder that use the old classes:
- plot_OTDA_classes.py
- plot_OTDA_color_images.py
- plot_OTDA_mapping.py
- plot_OTDA_mapping_color_images.py

We don't want the examples to be deprecated ;) Once this is done I will take care of updating the notebooks which is not done automatically yet.

About your questions, there should be a log parameter in init of the classes (note that it's log and not logs in all the toolbox as a parameter, log_ should be ok for storing the log ) and not the fit. Depending on this parameter the fit method should update the log_ parameter in the class. This parameter should exist all the time and be a an empty dictionary when log is set to False (maybe it should be set in the __init__ of the Base class ?).

I understand that the varying size return in the algorithm is surprising but it leads to a very simple use of the functions while still allowing expert user to get more information if needed. Changing the behavior of the algorithm now will completely break the toolbox and again is not the object of this PR. 
The reason for this is that computing the log can be time consuming since more things have to be returned (and the user should be able to choose of not to compute it).

ncourty:
Hi all, entering the discussion after some holiday break :)
@rflamary, regarding the notebooks, since we decided to go full sphinx-gallery for the example, it may be the right tempo to consider re-writing them in the sphinx syntax ?

agramfort:
" in the sphinx syntax" you mean in plain python :)

Slasnista:
Hi @rflamary,

- there si no error anymore when log=True
- I created a `.log_` that is either filled or instantiated as an empty dictionary depending on whether log is set to `True`
- for methods without log parameter (emd, lpl1) I removed log from parameters in the `__init__`
- I added some test to check that `log_` is filled when log=True.

rflamary:
Hello everyone,

@ncourty  since you are back from holiday and full of energy maybe you could take care of the test file formating ;) ? As POT maintainer you should be able to push commits on the PR.

@Slasnista thank you, every thing looks great, I will do a proper code review shortly but I think we will be able to do a merge shortly.

Slasnista:
Done with the examples according to the new OTDA objects !

@rflamary you're welcome, I now look forward to seeing it merged ;-)

ot/deprecation.py
```diff
@@ -0,0 +1,103 @@
+"""
```
rflamary: Personally I think this short module is a bit overkill, maybe move the class and function to the utils.py module with the origin URL.

ot/da.py
```diff
@@ -10,21 +10,28 @@
 # License: MIT License
 
 import numpy as np
+import warnings
```
rflamary: is it still necessary? I think everything is in deprecated now.

ot/da.py
```diff
@@ -706,12 +763,19 @@ def normalizeM(self, norm):
             self.M = np.log(1 + np.log(1 + self.M))
 
 
+@deprecated("The class OTDA_sinkhorn is deprecated in 0.3.1 and will be"
```
rflamary: Love the decorator! Removed in 0.5 is short but probably a good thing since it will disappear from the examples shortly

examples/da/plot_otda_classes.py
```diff
@@ -0,0 +1,142 @@
+# -*- coding: utf-8 -*-
```
rflamary: You chose to move the da examples in a sub-folder do you think it is more clear or is it to be similar to sk-learn?  did you remove the old ones, every trace of the old class should disappear from the examples?

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+
+        # introspect the constructor arguments to find the model parameters
+        # to represent
+        init_signature = signature(init)
+        # Consider the constructor parameters excluding 'self'
+        parameters = [p for p in init_signature.parameters.values()
+                      if p.name != 'self' and p.kind != p.VAR_KEYWORD]
+        for p in parameters:
+            if p.kind == p.VAR_POSITIONAL:
+                raise RuntimeError("scikit-learn estimators should always "
```
rflamary: remove scikit-learn and say POT ;)

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+                **self.kwargs)
             if self.bias:
                 K = np.hstack((K, np.ones((x.shape[0], 1))))
             return K.dot(self.L)
         else:
             print("Warning, model not fitted yet, returning None")
             return None
+
+##############################################################################
+# proposal
```
rflamary: remove proposal, we are near convergence here

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+##############################################################################
+
+
+# adapted from sklearn
+
+class BaseEstimator(object):
+    """Base class for all estimators in scikit-learn
```
rflamary: add POt reference and a link to the original source in sk-learn

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+    -------
+    mu : array-like, shape (n_samples,)
+        The uniform distribution estimated from X
+    """
+
+    return unif(X.shape[0])
+
+
+class BaseTransport(BaseEstimator):
+
```
rflamary: Add documentation and list of necessary parameters in __init__  for the class to run its functions 
such as metric

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+    def fit(self, Xs=None, ys=None, Xt=None, yt=None):
```
rflamary: Since we give all None values to the input parameters, we should handle those in case the user just call the function with nothing.

When a necessary parameter is missing we should print a warning and not do anything.

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+        self.mu_s = self.distribution_estimation(Xs)
+        self.mu_t = self.distribution_estimation(Xt)
+
+        # store arrays of samples
+        self.Xs = Xs
+        self.Xt = Xt
+
+        return self
+
+    def fit_transform(self, Xs=None, ys=None, Xt=None, yt=None):
```
rflamary: SAme here with the necessary parameters, maybe we should coce a test function with an error message that lists the necessary parameters i'm thinking as 

```python

def check_params(**kwargs):
    OK=True
    for param in kwargs:
        if kwargs[kwargs] is None:
            OK=False

    if not OK:
        #Print the list of Necessary params in kwargs

    return OK


# the function can be called inside the classes as
if check_params(Xs=Xs,Xt=Xt):
    do the stuff

```
What is nice is that we can give different keywords params depending on the class (semi supervised  or else).

This function or at leats a clener version should be in utils.py

ot/da.py
```diff
@@ -843,10 +927,896 @@ def predict(self, x):
...
+        batch_size : int, optional (default=128)
+            The batch size for out of sample inverse transform
+
+        Returns
+        -------
+        transp_Xs : array-like, shape (n_source_samples, n_features)
+            The transport source samples.
+        """
+
+        if np.array_equal(self.Xs, Xs):
```
rflamary: ```self.Xs is Xs ``` will test if the array is the same object and avoid the multiple tests and computation in np_array_equal.

ot/deprecation.py
Slasnista: Ok I do that

ot/da.py
Slasnista: I'll check whether it's used in the script otherwise I remove it

ot/da.py
Slasnista: Ok good point

ot/da.py
Slasnista: oh yes, i remove all the scikit refs like this one

ot/da.py
Slasnista: I do not do that for BaseTransport since it has no `__init__` and no real parameters but I will check the other to not forget parameters

ot/da.py
Slasnista: Good remark, I will handle the possible exceptions

ot/da.py
Slasnista: ok

examples/da/plot_otda_classes.py
Slasnista: - I think sub-categories would clearer for examples and later on for the doc generation
- I did not remove the old ones, shoud I move them in the new sub-folder ? @agramfort : do you have some advice to deprecate examples / documention jointly with some code ?

ot/da.py
Slasnista: Oh actually I need into the `BaseEstimator` class.

I will move the `BaseEstimator` classe into `utils.py`. That would be more relevant

examples/da/plot_otda_classes.py
rflamary: I think we shoud remove them from the examples since it's what people look at when trying out the toolbox.

Slasnista: Ok I do that so

ot/da.py
rflamary: what i'm saying is that you code here functions fit and transform that require some parameters/properties to be set before calling ()hence that should be defined in __init__. you should express here which properties should be set in order for the function to work.

rflamary: this to allows new contributors to propose new classes with no headaches ;)

ot/da.py
rflamary: Sorry but this is not what you did in the last commit, you just added a condition, oi think the implementation above would be more future proof since it handles the warning/error message automatically, no need for else.

examples/da/plot_otda_classes.py
agramfort: yes remove deprecated code from all examples and documentation

examples/da/plot_otda_color_images.py
```diff
@@ -61,62 +50,72 @@ def mat2im(X, shape):
 idx1 = np.random.randint(X1.shape[0], size=(nb,))
 idx2 = np.random.randint(X2.shape[0], size=(nb,))
```
agramfort: use RandomState

examples/da/plot_otda_d2.py
```diff
@@ -0,0 +1,163 @@
...
+# Authors: Remi Flamary <remi.flamary@unice.fr>
+#          Stanislas Chambon <stan.chambon@gmail.com>
+#
+# License: MIT License
+
+import matplotlib.pylab as pl
+import ot
+
+# number of source and target points to generate
+ns = 150
```
agramfort: n_samples_source = 150
n_samples_target = 150

it makes your comment above unnecessary and the code more readable

examples/da/plot_otda_d2.py
```diff
@@ -0,0 +1,163 @@
...
+nt = 150
+
+Xs, ys = ot.datasets.get_data_classif('3gauss', ns)
+Xt, yt = ot.datasets.get_data_classif('3gauss2', nt)
+
+# Cost matrix
+M = ot.dist(Xs, Xt)
```
agramfort: if it's doing euclidian distance than naming / API is not explicit enough IMO

examples/da/plot_otda_mapping.py
```diff
@@ -0,0 +1,119 @@
...
+
+np.random.seed(0)
+
+##############################################################################
+# generate
+##############################################################################
+
+n = 100  # nb samples in source and target datasets
+theta = 2 * np.pi / 20
+nz = 0.1
```
agramfort: nz ? could you use more explicit names?

examples/da/plot_otda_mapping.py
```diff
@@ -0,0 +1,119 @@
...
+Xt = Xt + 4
+
+
+# MappingTransport with linear kernel
+ot_mapping_linear = ot.da.MappingTransport(
+    kernel="linear", mu=1e0, eta=1e-8, bias=True,
+    max_iter=20, verbose=True)
+
+ot_mapping_linear.fit(
+    Xs=Xs, Xt=Xt)
```
agramfort: it fits on one line

examples/da/plot_otda_color_images.py
Slasnista: You mean a seed ?

examples/da/plot_otda_d2.py
rflamary: dist is basically a wrapper around scipy cdist where you can choose the metric.

Also it shouldn't be computed here since it's not used anywhere it seems.

Slasnista: Actually I did compute it to plot it along with source and target points

Slasnista:
Final modifications according to @agramfort comments pushed

examples/da/plot_otda_classes.py
```diff
@@ -15,8 +15,10 @@
 # License: MIT License
 
 import matplotlib.pylab as pl
+import numpy as np
 import ot
 
+np.random.seed(42)
```
rflamary: Actually I think @agramfort  prefer to use RandomSate as explained in
https://stackoverflow.com/questions/22994423/difference-between-np-random-seed-and-np-random-randomstate

This allow you to use controlable random generator without impacting the global one as is done in random.seed.

Note that we plan on doing proper dataset functions that accept a random state generator as parameter to allow full control (See #21 ).

Slasnista: no worries ;)

rflamary:
OK it is merged, thank you to @ngayraud @Slasnista  and @agramfort  for the work.

PR #25, opened 2017-08-29
Changes to LP solver:

aje:
- Allow to modify the maximal number of iterations
 - Display an error message in the python console if the solver encountered an issue

### DIFF
--- a/ot/da.py
+++ b/ot/da.py
@@ -13,7 +13,7 @@ import numpy as np
 
 from .bregman import sinkhorn
 from .lp import emd
-from .utils import unif, dist, kernel
+from .utils import unif, dist, kernel, cost_normalization
 from .utils import check_params, deprecated, BaseEstimator
 from .optim import cg
 from .optim import gcg
@@ -673,7 +673,7 @@ class OTDA(object):
         self.wt = wt
 
         self.M = dist(xs, xt, metric=self.metric)
-        self.normalizeM(norm)
+        self.M = cost_normalization(self.M, norm)
         self.G = emd(ws, wt, self.M, max_iter)
         self.computed = True
 
@@ -741,26 +741,6 @@ class OTDA(object):
         # aply the delta to the interpolation
         return xf[idx, :] + x - x0[idx, :]
 
-    def normalizeM(self, norm):
-        """ Apply normalization to the loss matrix
-
-
-        Parameters
-        ----------
-        norm : str
-            type of normalization from 'median','max','log','loglog'
-
-        """
-
-        if norm == "median":
-            self.M /= float(np.median(self.M))
-        elif norm == "max":
-            self.M /= float(np.max(self.M))
-        elif norm == "log":
-            self.M = np.log(1 + self.M)
-        elif norm == "loglog":
-            self.M = np.log(1 + np.log(1 + self.M))
-
 
 @deprecated("The class OTDA_sinkhorn is deprecated in 0.3.1 and will be"
             " removed in 0.5 \nUse class SinkhornTransport instead.")
@@ -787,7 +767,7 @@ class OTDA_sinkhorn(OTDA):
         self.wt = wt
 
         self.M = dist(xs, xt, metric=self.metric)
-        self.normalizeM(norm)
+        self.M = cost_normalization(self.M, norm)
         self.G = sinkhorn(ws, wt, self.M, reg, **kwargs)
         self.computed = True
 
@@ -816,7 +796,7 @@ class OTDA_lpl1(OTDA):
         self.wt = wt
 
         self.M = dist(xs, xt, metric=self.metric)
-        self.normalizeM(norm)
+        self.M = cost_normalization(self.M, norm)
         self.G = sinkhorn_lpl1_mm(ws, ys, wt, self.M, reg, eta, **kwargs)
         self.computed = True
 
@@ -845,7 +825,7 @@ class OTDA_l1l2(OTDA):
         self.wt = wt
 
         self.M = dist(xs, xt, metric=self.metric)
-        self.normalizeM(norm)
+        self.M = cost_normalization(self.M, norm)
         self.G = sinkhorn_l1l2_gl(ws, ys, wt, self.M, reg, eta, **kwargs)
         self.computed = True
 
@@ -1001,7 +981,7 @@ class BaseTransport(BaseEstimator):
 
             # pairwise distance
             self.cost_ = dist(Xs, Xt, metric=self.metric)
-            self.normalizeCost_(self.norm)
+            self.cost_ = cost_normalization(self.cost_, self.norm)
 
             if (ys is not None) and (yt is not None):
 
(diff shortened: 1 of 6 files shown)


ot/da.py
```diff
@@ -658,7 +658,7 @@ def __init__(self, metric='sqeuclidean'):
         self.metric = metric
         self.computed = False
 
-    def fit(self, xs, xt, ws=None, wt=None, norm=None):
+    def fit(self, xs, xt, ws=None, wt=None, norm=None, numItermax=10000):
```
rflamary: the default value for numItermax should be 100000 because it allows out of the box to handle larger matrices. Same for emd and emd2.

ot/lp/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 
 
-def emd(a, b, M):
+def emd(a, b, M, numItermax=10000):
```
rflamary: default value to 100000

ot/lp/__init__.py
```diff
@@ -84,9 +86,9 @@ def emd(a, b, M):
     if len(b) == 0:
         b = np.ones((M.shape[1], ), dtype=np.float64)/M.shape[1]
 
-    return emd_c(a, b, M)
+    return emd_c(a, b, M, numItermax)
 
-def emd2(a, b, M,processes=multiprocessing.cpu_count()):
+def emd2(a, b, M, processes=multiprocessing.cpu_count(), numItermax=10000):
```
rflamary: same here

Slasnista:
Hi @aje, according to @rflamary remark about OTDA object, you can update the class `EMDTransport()` as follows using `max_iter` var name to stick to the terminology introduced in PR #22 

```python
class EMDTransport(BaseTransport):
    """Domain Adapatation OT method based on Earth Mover's Distance

    Parameters
    ----------
    mapping : string, optional (default="barycentric")
        The kind of mapping to apply to transport samples from a domain into
        another one.
        if "barycentric" only the samples used to estimate the coupling can
        be transported from a domain to another one.
    metric : string, optional (default="sqeuclidean")
        The ground metric for the Wasserstein problem
    distribution : string, optional (default="uniform")
        The kind of distribution estimation to employ
    verbose : int, optional (default=0)
        Controls the verbosity of the optimization algorithm
    log : int, optional (default=0)
        Controls the logs of the optimization algorithm
    limit_max: float, optional (default=10)
        Controls the semi supervised mode. Transport between labeled source
        and target samples of different classes will exhibit an infinite cost
        (10 times the maximum value of the cost matrix)
    max_iter : int, float, optional (default=10000)
        The minimum number of iteration before stopping the optimization
        algorithm if no it has not converged

    Attributes
    ----------
    coupling_ : array-like, shape (n_source_samples, n_target_samples)
        The optimal coupling

    References
    ----------
    .. [1] N. Courty; R. Flamary; D. Tuia; A. Rakotomamonjy,
           "Optimal Transport for Domain Adaptation," in IEEE Transactions
           on Pattern Analysis and Machine Intelligence , vol.PP, no.99, pp.1-1
    """

    def __init__(self, metric="sqeuclidean",
                 distribution_estimation=distribution_estimation_uniform,
                 out_of_sample_map='ferradans', limit_max=10, max_iter=10000):

        self.metric = metric
        self.limit_max = limit_max
        self.distribution_estimation = distribution_estimation
        self.out_of_sample_map = out_of_sample_map
        self.max_iter = max_iter

    def fit(self, Xs, ys=None, Xt=None, yt=None):
        """Build a coupling matrix from source and target sets of samples
        (Xs, ys) and (Xt, yt)

        Parameters
        ----------
        Xs : array-like, shape (n_source_samples, n_features)
            The training input samples.
        ys : array-like, shape (n_source_samples,)
            The class labels
        Xt : array-like, shape (n_target_samples, n_features)
            The training input samples.
        yt : array-like, shape (n_labeled_target_samples,)
            The class labels

        Returns
        -------
        self : object
            Returns self.
        """

        super(EMDTransport, self).fit(Xs, ys, Xt, yt)

        # coupling estimation
        self.coupling_ = emd(
            a=self.mu_s, b=self.mu_t, M=self.cost_, numItermax=self.max_iter
        )

        return self
```

rflamary:
OK with @Slasnista , just use max_iter=100000 to allow for larger matrices out of the box.

I agree it is a better name but I want to keep consistency with the functional solvers for the moment. I think we should cleanup the function parameter names in 0.5 with deprecation warning anyway.

rflamary:
Ok Sorry @aje  if i wasn't clear but I want to keep numItermax for the emd and emd2 function as is used in the sinkhorn and sinkhorn2 for the moment. I'm Ok with max_iter in the classes though since we will switch to it in a next release.

rflamary:
Test is failing due to small naming error. I think when this is corrected we can do a quick merge.

ot/da.py
```diff
@@ -1182,6 +1183,26 @@ def inverse_transform(self, Xs=None, ys=None, Xt=None, yt=None,
 
             return transp_Xt
 
+    def normalizeCost_(self, norm):
```
rflamary: I like the normalization part but think this function should be independent from the class since it is of potential interest to any user.

I would define this function in utils.py with proper documentation. Then you can refer to this function in the doc of the classe

```python

def cost_normalization(C,norm=None):

    if norm.lower() == "median":
        C/=float(np.median(C))
    elif norm.lower() == "max":
        C/=float(np.max(C))
    elif norm.lower() == "log":
        C=np.log(1+C)
    elif norm.lower() == "loglog":
        C=np.log(1+np.log(1+C))

    return C
```

ot/da.py
```diff
@@ -1001,6 +1001,7 @@ def fit(self, Xs=None, ys=None, Xt=None, yt=None):
 
             # pairwise distance
             self.cost_ = dist(Xs, Xt, metric=self.metric)
+            self.normalizeCost_(self.norm)
```
rflamary: use an extern function as discussed below

ot/da.py
```diff
@@ -845,7 +825,7 @@ def fit(self, xs, ys, xt, reg=1, eta=1, ws=None, wt=None, norm=None,
         self.wt = wt
 
         self.M = dist(xs, xt, metric=self.metric)
-        self.normalizeM(norm)
+        self.M = cost_normalization(self.M, norm)
```
Slasnista: - @aje : can you handle the case where one does not want to normalise the cost matrix ?
- should we do that with a `if` in the `.fit` or in the `cost_normalization` function ?

rflamary: If you give it None is is by default the function will already return C. 

Seems OK to me.

aje: I could add the check
```
if self.norm is not None:
```
but this also seems ok to me

Slasnista:
ok, could you add the parameter `norm` in the `__init__`s of the different OTDA classes (except `BaseTransport()` ) and put None as default value ?

Slasnista:
or another default value if one normalization appears better that `None` ;)

rflamary:
@aje  I agree with @Slasnista  , the norm parameter should be defined in the ```__init__``` for the new classes. I didn't see that it was defined in fit.

@Slasnista  Let's keep it at None for the moment since there is not rule aboiut the best one and it has sometimes weird effect with regularized OT.

rflamary:
@Slasnista  could you do a second quick code review before we merge the PR?

Slasnista:
Ok from my side, everything seems good

Issue #26, opened 2017-09-01
semi supervised da - correction + example


Slasnista:
Hi, 

I may have found a line that could lead to errors when using OT objects in a supervised DA setting.

at line 992 in `da.py` I propose to change `classes = np.unique(ys)` into `classes = [c for c in np.unique(ys) if c != -1]` which would enable people to use source samples with no labels to find the optimal couplin.

I also propose to add an example for semi supervised DA.

Do you agree with these propositions ? If yes, I'll open a PR.

rflamary:
Hello @Slasnista 

OK but be very clear in the documentation about what -1 means because this could introduce bugs for people who have  2 classes and use -1,1 as class.

PR #28, opened 2017-09-04
Domain adaptation corrections

Slasnista:
cf issue #26 

- added an example for semi supervised DA
- correction for semi supervised DA in `.fit`
- added doc on semi supervised DA

### DIFF
--- a/examples/da/plot_otda_semi_supervised.py
+++ b/examples/da/plot_otda_semi_supervised.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+"""
+============================================
+OTDA unsupervised vs semi-supervised setting
+============================================
+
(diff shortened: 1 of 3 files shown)


Slasnista:
@rflamary oh indeed very good point. I do that.

Slasnista:
@rflamary : would it be worth planing a new pr to make semi supervised DA available also for `MappingTransport()` ? I guess that would widen the scope of application of this class.

Yet, since the cost matrix is not directly exhibited in this object, this would require some changes inside the mapping functions directly. Would you have any idea on how to proceed ?

rflamary:
Hello @Slasnista , I don't see any obvious and simple way to handle semi supervised with MappingTransport. 

If you find a way maybe try to publish it before sharing ;).

I will do the merge now.

PR #29, opened 2017-09-05
Dual variables in EMD_wrapper

arolet:
The main goal of this pull request is to let emd() return the dual variables of the linear program.

Dual variables are often useful in optimal transport because they are subgradients with respect to the input histograms. The lp solver is a dual simplex and already computes them, so returning them adds only a negligible cost.

There were some matlab function calls in the debugging parts of the lp solver (removed by the preprocessor when DEBUG_LEVEL==0), which were replaced with c-style prints.


rflamary:
Thank you @arolet this is a very interesting PR. The dual variables are indeed necessary in numerous applications.

It seems to run on travis now so that's perfect. I just think that emd and emd2 shoukd have a similar behavior than sinkhorn, i.,e. return a log (a dictionary) with the dual variables inside when parameter log is set to True.

arolet:
I added a log option which returns a dictionary. I also removed one of the two functions in emd_wrap.pyx because they only differed in their outputs and they do not seem to be meant to be called outside of lp/__init__.py

Also, previously when max_iter was reached it gave a message saying the problem is infeasible, which should not be the case I think. So I added feasibility checks before running the LP (the problem is feasible if both input are in the simplex), and I added a return value showing that max_iter was reached. Both cases trigger warnings.

ot/lp/emd_wrap.pyx
```diff
@@ -135,17 +77,13 @@ def emd2_c( np.ndarray[double, ndim=1, mode="c"] a,np.ndarray[double, ndim=1, mo
...
-    for i in range(n1):
-        for j in range(n2):
-            cost+=G[i,j]*M[i,j]
-
-    return cost
+            warnings.warn("Problem unbounded")
+        elif resultSolver == MAX_ITER_REACHED:
+            warnings.warn("numItermax reached before optimality. Try to increase numItermax.")
 
+    return G, cost, alpha, beta
```
rflamary: could you also return the status integer so that on can detect from the log  a convergence problem instead of just printing the error?

It should be added also to the log I think

ot/lp/__init__.py
```diff
@@ -94,12 +94,15 @@ def emd(a, b, M, numItermax=100000, log=False):
     if len(b) == 0:
         b = np.ones((M.shape[1],), dtype=np.float64) / M.shape[1]
 
-    G, cost, u, v = emd_c(a, b, M, numItermax)
+    G, cost, u, v, resultCode = emd_c(a, b, M, numItermax)
+    resultCodeString = checkResult(resultCode)
```
agramfort: you need to choose your side. In Python variables are snake_case not CamelCase. To be consistent with recent changes in package you should write:

result_code_string = check_result(result_code)

I know there is the C++ underneath that use CamelCase but in the binding you should adopt the conventions of the rest of the package.

my 2c

arolet: Right, thank you I didn't notice. I also replaced numItermax with num_iter_max and applied the same changes to the cython wrapper.

arolet:
Oh yes, it will also make it consistent with the name in the c++ code.

ot/lp/emd_wrap.pyx
```diff
@@ -12,81 +12,33 @@ cimport numpy as np
...
-             \gamma^T 1= b
-
-             \gamma\geq 0
-    where :
+    if result_code == INFEASIBLE:
+        message = "Problem infeasible. Check that a and b are in the simplex"
+    elif result_code == UNBOUNDED:
+        message = "Problem unbounded"
+    elif result_code == MAX_ITER_REACHED:
+        message = "numItermax reached before optimality. Try to increase numItermax."
```
agramfort: numItermax

git grep numItermax

is your friend.

arolet: Thank you, I'm rather new to git, I didn't know this one. My ide is supposed to rename variable names in strings but I guess it doesn't work well with cython...

It seems that there are a lot of variables called numIterMax across the repository, though I don't think the renaming of all of them should be done in this PR.

rflamary: Hello

Yes we have some renaming to do but it is not the object of this pr.

I'm ok with cleaning names in emd-c but the EMD and emd2 function should stick to the old parameter names because even if they do not follow convention they are consistent across the toolbox.

We will open an issue/or to discuss renaming and proper deprecation procedure for parameters. I would like to merge the current  price and release 0.3 shortly.

arolet: I reverted max_iter to its old name for now.

agramfort:
fine with me. One thing at a time :)

ot/lp/__init__.py
```diff
@@ -7,14 +7,16 @@
...
+
 # import compiled emd
-from .emd_wrap import emd_c, emd2_c
+from .emd_wrap import emd_c, check_result
 from ..utils import parmap
-import multiprocessing
 
 
-def emd(a, b, M, numItermax=100000):
+def emd(a, b, M, num_iter_max=100000, log=False):
```
rflamary: keep numItermax here o that we keep consistency with the rest of the toolbox

ot/lp/__init__.py
```diff
@@ -156,17 +177,30 @@ def emd2(a, b, M, processes=multiprocessing.cpu_count(), numItermax=100000):
...
+        b = np.ones((M.shape[1],), dtype=np.float64) / M.shape[1]
 
-    if len(b.shape) == 1:
-        return emd2_c(a, b, M, numItermax)
+    if log:
+        def f(b):
+            G, cost, u, v, resultCode = emd_c(a, b, M, num_iter_max)
+            result_code_string = check_result(resultCode)
+            log = {}
+            log['G'] = G
```
rflamary: i would remove G here because it does not scale well in memory when computing the distance between a lot of samples .

Maybe add a parameter return_matrix to the function?

ot/da.py
```diff
@@ -1370,7 +1370,7 @@ class label
 
         # coupling estimation
         self.coupling_ = emd(
-            a=self.mu_s, b=self.mu_t, M=self.cost_, numItermax=self.max_iter
+            a=self.mu_s, b=self.mu_t, M=self.cost_, num_iter_max=self.max_iter
```
rflamary: numItermax

arolet:
I thought I reverted to the original name but I made a mistake it seems. Should be fine now.

The matrix G is optionally put in the log now in emd2. If return_matrix is True, the value of log is ignored and all the optional return variables are put in the dictionary because they don't use nearly as much memory as G anyways.

PR #32, opened 2017-09-13
GPU changes:

toto6:
- Replace cudamat by cupy for GPU implementations (cupy is still in active development, while cudamat is not)
 - Use the new DA class instead of the old deprecated one

TODO for another PR:
 - Performances are still a bit lower than with cudamat (even if better than CPU for large matrices). Some speedups should be possible by tweaking the code


Slasnista:
hi @aje,

I have seen that you have just updated the gpu code and I look forward to using it !

Here some suggestions about gpu computing for maybe a next PR:
1. What do you think about merging the gpu.da and the standard da scripts. We could add just a boolean parameter in the init of each OTDA class to place the computations onto the GPU.

2. Along with this direction, what about having the same orga of modules, functions over the repo, classes... and use a boolean parameter for each function to specify whether the computation should be run on the gpu or on the cpu as for deep learning libraries (where you specify the device to use: tensorflow, theano, pytorch rely on this)

Let me know if you want to talk about it directly. We could plan a skype / hangout ;)

ncourty:
Hi @aje and @Slasnista, thanks for your work on POT ! It seems to me that the direction pointed out by @Slasnista is the good one for GPU integration in POT: seamless integration with a boolean that either can be set upon installation  or triggered by a test on the availability of cupy. However, I think it would be better to do this in another PR.

aje:
I agree it would be better to have a boolean than the current version where I directly duplicate code. I'm not sure how it could be done as I'm not familiar with other python library using GPU. For sure we would do a skype about it

rflamary:
Hello

The way i see it it can be solved rather elegantly with the following  functions in utils.py

```python

try:
    import cupy as cp
except ImportError:
    cp=False
    

def get_array_module(*args):
    """ return the proper modul (numpy of cupy) depending on the  """
    if cp:
        return cp.get_array_module(*args)
    else:
        return np
    
def to_gpu(*args):
    """ upload numpy arry to gpu and return them """
    return (cp.asarray(x) for x in args)
            
def is_gpu(np):
    """ test if a module is cupy on numpy"""
    if 'cupy' in np.__name__:
        return True
    else:
        return False

```
get_array_module return the module that should be used (numpy or cupy) depending on if its installed and the arrays you given them are already on gpu.

The second function transfer all the array given to it to the gpu. 

The last one allows for a quick test to know whether a module is numpy or cupy.

Now you just need to add at the beginning of any function (for isntance sinkhorn knopp):

```python

def sinkhorn_knopp(a, b, M, reg, numItermax=1000, stopThr=1e-9, verbose=False, log=False, **kwargs):

    np=get_array_module(a,b,M)
    
    a = np.asarray(a, dtype=np.float64) # probably to change depending on is_gpu(np)
    b = np.asarray(b, dtype=np.float64)
    M = np.asarray(M, dtype=np.float64)

```

This will automatically replace the local np  and use either numpy of cupy .

This means that obviously the user has to take care of sending the arrays on the gpu or indeed we add a parameter force_gpu that will force its use when cupy is installed 

```python
if force_cpu:
    a,b,M=to_gpu(a,b,M)
np=get_array_module(a,b,M)

```

This is the main strength of cupy  ( being mostly compatible with numpy) and we should do that instead of re-implementing everything.

Finally note that my proposition do not bug when cupy is not installed, it will live hapilly using numpy locally.

Slasnista:
I really like your solution @rflamary. It should work with every function, shouldn't it ?

We might be able to do this with a decorator applied to any functions. Then we would just have to make sure that the tests pass on gpu too. Does it sound tractable or am I forgetting some issue ?

aje:
It seems like a good solution to me. I will try to make it work.

rflamary:
@Slasnista I don't know about the decorator, how do you handle the local np with it?

I think indeed it should work with every functions that use the numpy subset implemented in cupy.

@aje I think we should also handle the case when cupy is bot installed in to_gpu: 

```python
def to_gpu(*args):
    """ upload numpy arry to gpu and return them """
    if cp:
        return (cp.asarray(x) for x in args)
    else:
        return args
```
now we have a code that does not explode with force_gpu , maybe we should add a  warning to tell the user that cupy is not installed and that default backend is used....

Slasnista:
@rflamary, a decorator `to_gpu` would allow you to do something like this to handle the conversion cpu <-> gpu:

```python
@to_gpu
def ot_function(X, y, reg=1., verbose=False):
    """ot_function
    """

     ...

     return Coupling
```

If the decorator is well designed he may be able to turn any numpy array given as argument into a cupy array and also transform the returned cupy array into a numpy array after the function has run.

After that, we would just have to put this decorator on every function that performs a heavy computation that would be speeded up onto a gpu. Does it sound good to you ?

rflamary:
Hello,
@Slasnista 
+ I agree that a decorator is both elegant and the way to go for handling upload to gpu. syill I dont' want to convert back to numpy array since it introduce and overhead every time you can the function that the user might not need. My opinion is that if the user wants to use the gpu, he can handle gpu array in the output. Still one could add a parameter to_array=False that says if the arrayx should be converted automatically.
+ the decorator can handle only array conversion since we still need to get the local backend inside the function so it does to solve all the modifications .


@aje 
+ I'm not convince with the use of xp instead of np. This make the function non numpy standard and if a user wants to use the function in some other place by removing the xp=get... line it would force him to import numpy as xp (highly irregular). I also think it is a good idea to keep np to keep the code easy to read for outsiders maybe a simple comment above the line can say ("overwrite backend with numpy of cupy depending on array format").
+ could you try to code the decorator for loading the arryy onto gpu instead of handling it inside the function as propose by @Slasnista ? If not maybe @Slasnista can propose an implementation.

rflamary:
Also I think we should have default dtytes for GPU and CPU, CPU can handle float64 very easily but GPU has a very large overhead (which explains the limited performance gain of cupy for the moment). i' pretty sure that using float32 would lead to tremendous computational gain (at the cost of less numerical stability obviously)

Maybe we should let the user define the dtype through the array passed and not force it inside the sinkhorn function.

aje:
I don't understand this decorator thing, I never used any

aje:
Concerning the overhead with float64 or float32, I think this is not the main cause of bad performances here.

I remember, with cudamat, I had significantly better performances (and with float64). The performances with cudamat where the following:
 5000
Normal sinkhorn, time:   1.21 sec
min:5.251E-12 max::8.839E-05 mean::1.600E-07 std::5.317E-07
   GPU sinkhorn, time:   0.09 sec
min:5.251E-12 max::8.839E-05 mean::1.600E-07 std::5.317E-07

 10000
Normal sinkhorn, time:   4.37 sec
min:4.508E-13 max::4.601E-05 mean::4.000E-08 std::1.366E-07
   GPU sinkhorn, time:   0.25 sec
min:4.508E-13 max::4.601E-05 mean::4.000E-08 std::1.366E-07

 15000
Normal sinkhorn, time:  10.68 sec
min:1.696E-13 max::2.031E-05 mean::1.778E-08 std::6.191E-08
   GPU sinkhorn, time:   0.49 sec
min:1.696E-13 max::2.031E-05 mean::1.778E-08 std::6.191E-08

 20000
Normal sinkhorn, time:  22.55 sec
min:1.144E-13 max::1.383E-05 mean::1.000E-08 std::3.506E-08
   GPU sinkhorn, time:   0.89 sec
min:1.144E-13 max::1.383E-05 mean::1.000E-08 std::3.506E-08

 25000
Normal sinkhorn, time:  32.90 sec
min:8.452E-14 max::8.882E-06 mean::6.400E-09 std::2.248E-08
   GPU sinkhorn, time:   1.39 sec
min:8.452E-14 max::8.882E-06 mean::6.400E-09 std::2.248E-08

 30000
Normal sinkhorn, time:  47.70 sec
min:6.306E-14 max::9.280E-06 mean::4.444E-09 std::1.572E-08
   GPU sinkhorn, time:   1.93 sec
min:6.306E-14 max::9.280E-06 mean::4.444E-09 std::1.572E-08

rflamary:
@aje I agree with you this difference in performance is clearly a problem and i'm disappointed by the performances of cupy so far. I will try it on my machien to see if the gain is still so small on a different GPU.

rflamary:
hello @aje 

I just tested a bit pycu with the following script

```python
import numpy as np
import matplotlib.pylab as pl
import cupy as cp
import ot

#%% parameters and data generation
n=5000 # nb samples
tp=np.float64

A=np.random.randn(n,n).astype(tp)
B=np.random.randn(n,n).astype(tp)

A_gpu=cp.asarray(A)
B_gpu=cp.asarray(B)

ot.tic()
A.dot(B)
ot.toc()

ot.tic()
A_gpu.dot(B_gpu)
ot.toc()
```

and got the following performances for float32:

```
Elapsed time : 0.601469993591 s
Elapsed time : 0.0522150993347 s
```

and for float64:

```
Elapsed time : 1.05664110184 s
Elapsed time : 1.37450909615 s
```

I've got a Titan X that is a bit old and sucks at float64 but I think it comes mainly from the fact that cupy is part of chainer that is a deep learning framework and they simply don't care about float64 (they even talk about going 16 of 8 bit in neural networks nowadays).

aje:
ah, I have similar results. I will try to add a parameter to chose which float type (32 or 64) to use in the computation

rflamary:
I think the type of the array given to the function should be the type of the array used in the gpu

Let the user select the float type outside is probably better because else we will have to add a dtype parameter to all gpu functions...

just use asarray with no d_type and it should work, obviously we will need to explain it in the documentation ;).

aje:
I did this change and I modified my little benchmark script to use both float64 and float32. The results are the following:
```
Using <class 'numpy.float32'>
        scipy's cdist, time:  11.18 sec 
pairwiseEuclidean CPU, time:   0.32 sec 
pairwiseEuclidean GPU, time:   0.09 sec 
Sinkhorn CPU, time:   2.14 sec 
Sinkhorn GPU, time:   0.99 sec 
Using <class 'numpy.float64'>
        scipy's cdist, time:  11.62 sec 
pairwiseEuclidean CPU, time:   0.73 sec 
pairwiseEuclidean GPU, time:   0.34 sec 
Sinkhorn CPU, time:   3.74 sec 
Sinkhorn GPU, time:   2.66 sec 
```
Still better with float32. The gain do not seem as large as your example script, so I might have missed some things

rflamary:
Thank you @aje 

I'm indeed not conviced by the very smal computationnal gain.

maybe some operations are not performed well  with cupy. We have to investigate that. I will try your branch and have a look with a profiler. Feel free to do the same on your end and we can maybe find the problem.

rflamary:
Found one bug!

you have to make sinkhorn gpu proof because right now you call sinkhorn in the class and you xan see line  230 of bregman.py that there is a np.asarray that convert everything back to numpy array.

I solved it temporarily by calling directly sinkhorm-knop in the domain adaptation Class and got far better results.

Before I had the following performances:
```
Using <type 'numpy.float32'>
        scipy's cdist, time:   7.41 sec 
pairwiseEuclidean CPU, time:   0.51 sec 
pairwiseEuclidean GPU, time:   6.28 sec 
Sinkhorn CPU, time:   4.32 sec 
Sinkhorn GPU, time:  30.60 sec 
Using <type 'numpy.float64'>
        scipy's cdist, time:   7.31 sec 
pairwiseEuclidean CPU, time:   0.90 sec 
pairwiseEuclidean GPU, time:   2.80 sec 
Sinkhorn CPU, time:  22.60 sec 
Sinkhorn GPU, time:  23.37 sec 
```

that were so weird and now I have this:
```
Using <type 'numpy.float32'>
        scipy's cdist, time:   7.41 sec 
pairwiseEuclidean CPU, time:   0.57 sec 
pairwiseEuclidean GPU, time:   0.03 sec 
Sinkhorn CPU, time:   5.10 sec 
Sinkhorn GPU, time:   0.31 sec 
Using <type 'numpy.float64'>
        scipy's cdist, time:   7.51 sec 
pairwiseEuclidean CPU, time:   0.94 sec 
pairwiseEuclidean GPU, time:   0.16 sec 
Sinkhorn CPU, time:  11.78 sec 
Sinkhorn GPU, time:   1.18 sec 
```

It's not a big gain but it's now definitely better.

aje:
hmmm strange, because line 230 in bregman.py is part of function "sinkhorn2". But it should not be called because the domain adaptation Class use the function "sinkhorn".
Another strange thing in your performance gain is that this is also for the "pairwiseEuclidean GPU" that goes from 6.28s to 0.03s for float32. And this function only compute the pairwise matrix, so you didn't change the code for this, but you still have performance improvement

aje:
I'm not sure if this is related, but every time I execute a script that use GPU, the first GPU call is slower than the following ones: if I execute the same thing twice, the second time is faster. Maybe you also had this strange behavior ?

rflamary:
Hello @aje 

I agree with you that this line 230 is not a problem.

After a reboot I get this for a cold start
```
Using <type 'numpy.float32'>
        scipy's cdist, time:   7.41 sec 
pairwiseEuclidean CPU, time:   0.48 sec 
pairwiseEuclidean GPU, time:   1.18 sec 
Sinkhorn CPU, time:   3.96 sec 
Sinkhorn GPU, time:   0.54 sec 
Using <type 'numpy.float64'>
        scipy's cdist, time:   7.41 sec 
pairwiseEuclidean CPU, time:   0.88 sec 
pairwiseEuclidean GPU, time:   0.21 sec 
Sinkhorn CPU, time:  18.41 sec 
Sinkhorn GPU, time:   1.28 sec 
```

at this for a second run

```
Using <type 'numpy.float32'>
        scipy's cdist, time:   7.43 sec 
pairwiseEuclidean CPU, time:   0.46 sec 
pairwiseEuclidean GPU, time:   0.03 sec 
Sinkhorn CPU, time:   4.01 sec 
Sinkhorn GPU, time:   0.29 sec 
Using <type 'numpy.float64'>
        scipy's cdist, time:   7.40 sec 
pairwiseEuclidean CPU, time:   0.88 sec 
pairwiseEuclidean GPU, time:   0.15 sec 
Sinkhorn CPU, time:   7.29 sec 
Sinkhorn GPU, time:   1.17 sec 
```
I agree that the bench has to be run twice but still don't understand what 

On the second run we get a speedup of  x13 for float32 and  x6 for flaot64 that are reasonable at least.

aje:
Ok, I guess I have worse performances than you because of my GTX 980M.
I will add the GPU implementation for LpL1, remove the no longer used ot.gpu module, and then it should be good to merge

rflamary:
@aje before implementaing all thos things you should have a look at the  "deprecated" decorator in utils.py.

I think a decorator is still the good way to handle the to-array upload automatically.

I will try to push an implementation today so that you see what we mean by that.

aje:
Yes, an example could be good, because I'm not sure how it would work to upload on the gpu, knowing that we should give the possibility to the user to chose if yes or no he wants the computation to be done on the gpu, like I did for the function pairwiseEuclidean in file utils.py

rflamary:
OK @aje and @Slasnista, I  managed to code a decorator in utils.py in this branch cloned from your PR:
https://github.com/rflamary/POT/tree/gpu_cupy

basically around any numpy function you can do something like this

```python

@gpu_fun(in_arrays=[0,1,2],out_arrays=[0]) # the index refer to the arrays so that we know which one to upload/recover
def myfunction(x,y,z,autre_param=False):
    np = get_array_module(x,y,z)
    ...
   return M,log

# now you can call the function with
 M,log=myfunction(x,y,z,autre_param=False,gpu=True)
# it will automatically upload x,y and z to gpu  if you set gpu=True or not if gpu_false


# if you want the function to return numpy arry you call it like that:
 M,log= myfunction(x,y,z,autre_param=False,gpu=True,to_np=True)
# the M will be a numpy array
```

I applied it to you pairwise distance function (we need to change its name btw) and removed the gpu parameters since it is no longer needed. It has been applied also to sinkhorn and sinkhorn_knopp and literally took one copy paste to make then accept  gpu and to_np parameters!

I think this is nice because it allows the user to choose whether he wants to upload automatically if convert back automatically.

I think in any case this decorator should be part of the toolbox (you can use it un you script!) still i dont know if i want to decorate all the functions because it kind of breaks the function definition (you see only args and kwargs) and now you need the doc to know how to call the function. 

Also note that you still need to use the np = get_array_module(...) function inside the function so it is not really fully numpy compatible (just one line to change though...) but it avoids the additionnal gpu parameter.

Finally the deprecated decorator is a class but I don't think it is necessary here we can use a function as decorator for this rather simple wrapper.

Slasnista:
@rflamary great proposal, i like it !

Is it compatible with any function / classes ? For instance for the da class, could it be used easily to turn the coupling optimization onto the gpu and then back to the cpu ? If yes, we may make the suitable modifications in this pr too.

Is there a way to write unitary tests for the function that may run onto the gpu and the cpu and run them automatically with Jenkins ? This way we could check automatically first that the cpu / gpu computations give the same thing in small data case. Second we could also check that when a gpu device is available, everything goes right. Finally, when a gpu is not available everything is done on cpu without error. We should maybe write test cases for these situations, shouldn't we ?

rflamary:
Hello @aje  and @Slasnista,

The decorator works for the moment only on functions (any function that uses get_array_module  indeed) but the Class decorator can also be coded.

@Slasnista I like it also but I definitively don't have the time to work on it. It also obviously misses tests. For the moment I think when gpu is not installed everything works even with parameter gpu because we return np if cupy is not installed. But note that we don't do jenkins at the moment only travis and I don't think it can handle GPU (at leas for free). If one of you want to continue on this it's great but i won't be able to finish it properly before a few months i think.

aje:
I think I finished the PR. Any changes required before merge ?

rflamary:
Hello @aje 

We are currently at NIPS with @ncourty but we will try and do a final code review shortly. Thank you for all the work and making pot more future-proof.

rflamary:
Hello @aje 

Very sorry for the unresponsive PR but I am having a lot of difficulties with my GPU/driver so I'l unable to do a proper code review and test for the moment.

@ncourty  could you have a look at the PR please? It will be nice to have a new set of eyes anyway.

rflamary:
Hello everyone and sorry for the long wait.

I have been playing with this PR now that I have a working GPU again and you can see a few modifications here:
https://github.com/rflamary/POT/tree/pr32

I think it's nice and you get a computational gain on large matrices but i have a major problem with the effect of the decorator functions: they screw with the documentation and function signature.

Basically not autocomplete for the sinkhorn function gives you ```sinkhorn(*args,**kwarg)``` which is not good at all. Same with the documentation that do not have a proper signature anymore.

There is supposedly a module called ```decorator``` that provide this sort of thing but i have not been able to use it properly (keeps signature but function do not work).

This is a major problem because documentation is very important and we cannot have a good one on the functions that go through the decorators (signature is part of the doc). 

The only solution in my opinion (unless one of you manage to use the decorator module) is to provide separate functions  which  go though the decorators. They will not have a proper signature but at least the original ones will have proper signature and since you can use this function as a drop-in replacement it's OK.  
Those functions can be defined in  a separate gpu.py file or directly in ot.bregman. The last one would require only small changes from the OptimalTransport classes.

What do you think?

agramfort:
maybe it's worth looking how tensorly https://github.com/tensorly/tensorly
handles different backends without duplicating much algorithmic code.

my 2c

aje:
Since the decorator upload/download matrices to/from the GPU at the beginning/end of the functions that have the decorator, a possibility is to remove the decorator and add one line of code at the beginning/end of these functions to do the same thing (this line would call a function that upload/download if necessary). This way, there is no need to duplicate the functions. But it would require the functions having the decorator to receive the gpu/to_np arguments through kwargs.

But I will check the decorator module to see if I can make it work without removing the decorator.

rflamary:
Ok thank you.

Please merge my branch first. It will help you merge master I already handled most conflicts

aje:
I merged the branch.
The decorator module you speak of is this https://github.com/aje/POT/commit/90636c44d278de19ae68534bdb63003d9bde874d#diff-55ba92bcceef627e33e92d3440cc23fc ?

rflamary:
yes indeed i did notw ant to add a dependency and since the module is a unique .py file it seemed a good idea to put it in externals

rflamary:
Hello @toto6 ,

Thank you for your work on this PR. It was a big help when refactoring the ot.gpu.
But since it is now outdated it cannot be merged anymore and I will close it.

PR #36, opened 2017-10-27
remove automatic imports from ot/__init__.py

Slasnista:
This PR intends to remove all automatics imports when importing the library.

In particular, it solves crashes when pot is imported on a instance without connected display device.

### DIFF
--- a/ot/__init__.py
+++ b/ot/__init__.py
@@ -10,28 +10,28 @@
 # License: MIT License
 
 
-# All submodules and packages
-from . import lp
-from . import bregman
-from . import optim
-from . import utils
-from . import datasets
-from . import plot
-from . import da
-from . import gromov
+# # All submodules and packages
+# from . import lp
+# from . import bregman
+# from . import optim
+# from . import utils
+# from . import datasets
+# from . import plot
+# from . import da
+# from . import gromov
 
-# OT functions
-from .lp import emd, emd2
-from .bregman import sinkhorn, sinkhorn2, barycenter
-from .da import sinkhorn_lpl1_mm
-from .gromov import gromov_wasserstein, gromov_wasserstein2
+# # OT functions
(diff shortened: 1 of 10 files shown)


rflamary:
Hello @Slasnista,

Removing all the automatic imports is rather extreme and make the toolbox more difficult to use  IMHO. 

I agree that ot.plot force a matplotlib dependency AND will break on servers with no display so we should remove it but i definitively want to keeop all the others submodules import that require only numpy and scipy (since the toolbox needs then anyway).

Also this a a feature breaker for all the previous code and should be thoroughly discussed before any commitment with everybody  (@ncourty i'm looking at you too).

Maybe don't convert all the examples yet until we have reached an agreement.

agramfort:
I thought the purpose was just to remove matplotlib from automatic imports

I agree that all these changes are probably unecessary

Slasnista:
Hi @rflamary 

Actually, it does not change much things. Just instead of doing:
```python
import ot

otda = ot.da.SinkhornTransport()
```

we would do 

```python

from ot.da import SinkhornTransport

otda = SinkhornTransport()
```

Slasnista:
Otherwise I will juste remove the line concerning the import of the plot module when the library is imported. It would solve the problem without changing too many things

rflamary:
yes I would prefer just to remove the plot line in ```__init__.py```. 

I know that it does not change much things but the first code is how i work with pot so i would need to update all my OT code. Also you can still use the second one in your own implementation but not mine anymore ;).

agramfort:
>
> yes I would prefer just to remove the plot line in __init__.py.
>
+1

rflamary:
Hello @Slasnista 

I close this PR because as we discussed above we just need to remove the plot in ot imports and I did it in  #41.

Thank you for the Issue, you just made POT easier to use with no graphics.

PR #39, opened 2018-01-08
typo in title

DeVerMyst:
target instead of traget


rflamary:
Hello @DeVerMyst ,

The checks seem to fail because of a variable name that is too simple. this si old stuff that is not your fault but should be handled before merging.

I propose to remplce I by Id for the identity matrix and the test should pass afterward.

rflamary:
Hello @DeVerMyst,

Could you please merge your PR with the current master so that we can check that the build happens?

I think we will be able to merge after that, thank you

Rémi

PR #44, opened 2018-04-18
Using log parameter

tardyb:
I have added the use of log parameter in class EMDTransport and SinkhornLpL1Transport, where it was not used. Related discussion here: https://github.com/rflamary/POT/issues/43

### DIFF
--- a/ot/da.py
+++ b/ot/da.py
@@ -1332,13 +1332,14 @@ class EMDTransport(BaseTransport):
            on Pattern Analysis and Machine Intelligence , vol.PP, no.99, pp.1-1
     """
 
-    def __init__(self, metric="sqeuclidean", norm=None,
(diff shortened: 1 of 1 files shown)


rflamary:
Hello @tardyb and thank you for the PR,

Currently the build fails because of small pep8 problems. Note that we enforce strict pep8 in all the POT code which can be done automatically with autopep8

```
$ flake8 examples/ ot/ test/
ot/da.py:1335:55: E231 missing whitespace after ','
ot/da.py:1339:1: W293 blank line contains whitespace
ot/da.py:1494:37: E231 missing whitespace after ','
The command "flake8 examples/ ot/ test/" exited with 1.
```

but since you only have 3 problems they can be easily corrected directly in the code.

Once the code build I think we can quickly merge your PR.

PR #45, opened 2018-05-02
add markdown format for pipy

rflamary:
Add information for a markdown long description in pipy (which is now compatible).

This aim at avoiding this:
https://pypi.org/project/POT/0.4.0/

and getting something such as this:
https://pypi.org/project/POT/0.3.1/

The fail was due to  lack of pandoc on my machine at the time.


agramfort:
maybe:

python setup.py check --restructuredtext

can help.

PR #47, opened 2018-05-14
LP Wasserstein barycenter with scipy linear solver and/or cvxopt

rflamary:
The PR add the following:

+ LP solver for for Wassersein barycenters (do not scale at all but still nice for comparison).
+ Can choose the solver between scipy.optimize and cvxopt (the new interior point solver of scipy as pretty good)
+ Documentation and test for the solver, update of the README.
+ Example of comparison between LP and regularized barycenter :

image

### DIFF
--- a/ot/bregman.py
+++ b/ot/bregman.py
@@ -845,7 +845,7 @@ def barycenter(A, M, reg, weights=None, numItermax=1000,
     reg : float
         Regularization term >0
     weights : np.ndarray (n,)
-        Weights of each histogram i_i on the simplex        
+        Weights of each histogram i_i on the simplex
     numItermax : int, optional
         Max number of iterations
     stopThr : float, optional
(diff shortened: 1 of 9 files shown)


ot/bregman.py
```diff
@@ -844,6 +844,8 @@ def barycenter(A, M, reg, weights=None, numItermax=1000,
         loss matrix   for OT
     reg : float
         Regularization term >0
+    weights : np.ndarray (n,)
+        Weights of each histogram i_i on the simplex
```
ncourty: each histogram i_i -> each histogram bin i

ot/lp/cvx.py
```diff
@@ -0,0 +1,146 @@
...
+    Parameters
+    ----------
+    A : np.ndarray (d,n)
+        n training distributions of size d
+    M : np.ndarray (d,d)
+        loss matrix   for OT
+    reg : float
+        Regularization term >0
+    weights : np.ndarray (n,)
+        Weights of each histogram i_i on the simplex
```
ncourty: same here

PR #50, opened 2018-05-31
Smooth and Sparse OT

rflamary:
This PR add several solvers for regularized OT from the following paper:

 Blondel, M., Seguy, V., & Rolet, A. (2018). [Smooth and Sparse Optimal Transport](https://arxiv.org/abs/1710.06276). Proceedings of the Twenty-First International Conference on Artificial Intelligence and Statistics (AISTATS).

Some of the code come from https://github.com/mblondel/smooth-ot/

List of new stuff:
- Solvers for smooty dual and semi dual for negentropy and squared L2 regularization
- Wrapper functions `smooth_ot_dual` and `smooth_ot_semi_dual` compatible with ot.sinkhorn.
- Detailed dosumentation about the optimization problems.
- Several test for all regularizations.
- Example of use code in `plot_OT_1D_smooth.py`
- Update readme with new feature and reference.

Here is an example of 1D OT matrix for squared L2 regularization.
image


README.md
```diff
@@ -15,7 +15,8 @@ It provides the following solvers:
 
 * OT Network Flow solver for the linear program/ Earth Movers Distance [1].
 * Entropic regularization OT solver with Sinkhorn Knopp Algorithm [2] and stabilized version [9][10] with optional GPU implementation (requires cudamat).
-* Non regularized Wasserstein barycenters [16] with LP solver.
+* Smooth optimal transport solvers (dual and semi-dual) for KL and squared L2 regularization [17].
```
ncourty: regularization's'

examples/plot_OT_1D_smooth.py
```diff
@@ -0,0 +1,110 @@
...
+
+# Author: Remi Flamary <remi.flamary@unice.fr>
+#
+# License: MIT License
+
+import numpy as np
+import matplotlib.pylab as pl
+import ot
+import ot.plot
+from ot.datasets import get_1D_gauss as gauss
```
ncourty: make_1D_gauss now

ot/smooth.py
```diff
@@ -0,0 +1,593 @@
...
+        Parameters
+        ----------
+        gamma: float
+            Regularization parameter.
+            We recover unregularized OT when gamma -> 0.
+
+        """
+        self.gamma = gamma
+
+    def delta_Omega(X):
```
ncourty: should we keep those functions here ?

ncourty: All the ones that raise a 'not implemented' exception

rflamary: this is proper base class definition I believe, it provide  aclean base implementation with proper Errors in can of use. I will complete the documentation so that it is clear for a user that this class should not be used in practice.

mblondel:
Thanks a lot for the integration work in POT!

PR #52, opened 2018-06-06
add stochastic methods

kilianFatras:
SAG for discrete measures and ASGD for semi continuous measure.


rflamary:
Hello @kilianFatras  and thank you for the job. We needed the stochastic implementation and you provided a good implementation there. 

But there are  still a few things that need to be done before merging

+ The PR do not build because of pep8 problems and failing tests. 
+ The submodule stochastic is not imported in module ot which is why it does not work in the test. You need to add ```from . import stochastic``` to the ```ot/__init__.py``` file in order to import it while doing import ot.
+ The order of the parameters in all your function should follow closely the order in other functions such as ot.emd and ot.sinkhorn so that you can switch from one to the other easily (at least a,b M , and reg (which you call epsilon and should be renamed also) followed by optimization parameters). 
+ n_source and n_target are not useful since those dimensions can be inferred from a,b or M. You also should provide default values for n_iter (renamed to numItermax to fit ot.emd and ot.sinkhorn functions) and lr.
+ In the test i would like you to also test that your solver return the same solution as the ot.sinkkorn function. This is an important test since both solver address the same optimization problem with different algorithms.
+ please use POT functions for providing uniform distributions and computing distance matrices M (in the doc, the test and the examples). uniform distributions of dimensionality n are obtained easily by ot.unif, and M can be computed very easily with ot.dist function.
+ We try to provide in the documentation of the function the optimization problem that is solved (see ot.emd or ot.sinkhorn). So plesa express the dual problem solved by the saga and asgd function. The c transform should also appear in the documentation since it can be used in numerous occasions. Also try to provide the euaton or algorithm number of the reference paper use in the documentation of all the functions. It helps the user/reader to better  follow what is done.

Note that all the tests are performed again after you do a commit push.

Again thank you for the job, those are only remarks aiming at having a better integration of your algorithm into POT.

Rémi

ot/stochastic.py
```diff
@@ -0,0 +1,412 @@
...
+    n_source = np.shape(M)[0]
+    u = np.zeros(n_source)
+    for i in range(n_source):
+        r = M[i, :] - v
+        exp_v = np.exp(-r / reg) * b
+        u[i] = - reg * np.log(np.sum(exp_v))
+    return u
+
+
+def transportation_matrix_entropic(a, b, M, reg, method, numItermax=10000,
```
rflamary: rename the function to ```solve_entropic```

ot/stochastic.py
```diff
@@ -0,0 +1,412 @@
...
+                     Advances in Neural Information Processing Systems (2016),
+                      arXiv preprint arxiv:1605.08527.
+    '''
+
+    n_source = np.shape(M)[0]
+    u = np.zeros(n_source)
+    for i in range(n_source):
+        r = M[i, :] - v
+        exp_v = np.exp(-r / reg) * b
+        u[i] = - reg * np.log(np.sum(exp_v))
```
rflamary: should we use a stabilized version of logsumexp?
https://en.wikipedia.org/wiki/LogSumExp

so that the algorithm do not explode

ot/stochastic.py
rflamary: or ```solve_semi_dial_entropic```

test/test_stochastic.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # check constratints
+    np.testing.assert_allclose(
+        zero, (G_sag - G_sinkhorn).sum(1), atol=1e-03)  # cf convergence sag
+    np.testing.assert_allclose(
+        zero, (G_sag - G_sinkhorn).sum(0), atol=1e-03)  # cf convergence sag
+    np.testing.assert_allclose(
+        zero, (G_asgd - G_sinkhorn).sum(1), atol=1e-03)  # cf convergence asgd
+    np.testing.assert_allclose(
+        zero, (G_asgd - G_sinkhorn).sum(0), atol=1e-03)  # cf convergence asgd
```
rflamary: also test the whole matrix and not onmly the marginals  with ```np.testing.assert_allclose(G_asgd nG_sinkhorn, atol=1e-03)```

ot/stochastic.py
```diff
@@ -0,0 +1,412 @@
...
+    elif method.lower() == "asgd":
+        opt_v = averaged_sgd_entropic_transport(b, M, reg, numItermax, lr)
+    else:
+        print("Please, select your method between SAG and ASGD")
+        return None
+
+    opt_u = c_transform_entropic(b, M, reg, opt_v)
+    pi = (np.exp((opt_u[:, None] + opt_v[None, :] - M[:, :]) / reg) *
+          a[:, None] * b[None, :])
+    return pi
```
rflamary: Please add a log parameter to the function that return a dictionary with u and v (probably better name alpha and beta see sinkhorn/emd wher emd return also the dual)

examples/plot_stochastic.py
```diff
@@ -0,0 +1,205 @@
...
+# Call the "ASGD" method to find the transportation matrix in the semicontinous
+# case
+# ---------------------------------------------
+#
+# Define the method "ASGD", call ot.solve_semi_dual_entropic and plot the
+# results.
+
+method = "ASGD"
+asgd_pi, log = ot.stochastic.solve_semi_dual_entropic(a, b, M, reg, method,
+                                                      numItermax, log)
```
rflamary: bug because log is the wrong positional argument. Replace ```log``` by  ```log_asgd``` is the left hand side and ```log``` by  ```log=log``` in the right hand side.

examples/plot_stochastic.py
```diff
@@ -0,0 +1,205 @@
...
+
+"""
+
+# Author: Kilian Fatras <kilian.fatras@gmail.com>
+#
+# License: MIT License
+
+import matplotlib.pylab as pl
+import numpy as np
+import ot
```
rflamary: also  ```import ot.plot``` because it is not imported by default

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+# Optimization toolbox for SEMI - DUAL problems
+##############################################################################
+
+
+def coordinate_grad_semi_dual(b, M, reg, beta, i):
+    '''
+    Compute the coordinate gradient update for regularized discrete
+        distributions for (i, :)
+
+    The function computes the gradient of the semi dual problem:
```
rflamary: add line before math operator for proper math generation in documentation

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+    return b - khi
+
+
+def sag_entropic_transport(a, b, M, reg, numItermax=10000, lr=None):
+    '''
+    Compute the SAG algorithm to solve the regularized discrete measures
+        optimal transport max problem
+
+    The function solves the following optimization problem:
+    .. math::
```
rflamary: same here

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+    return cur_beta
+
+
+def averaged_sgd_entropic_transport(a, b, M, reg, numItermax=300000, lr=None):
+    '''
+    Compute the ASGD algorithm to solve the regularized semi contibous measures
+        optimal transport max problem
+
+    The function solves the following optimization problem:
+    .. math::
```
rflamary: same here

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+    return ave_beta
+
+
+def c_transform_entropic(b, M, reg, beta):
+    '''
+    The goal is to recover u from the c-transform.
+
+    The function computes the c_transform of a dual variable from the other
+    dual variable:
+    .. math::
```
rflamary: same here

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+
+
+def solve_semi_dual_entropic(a, b, M, reg, method, numItermax=10000, lr=None,
+                                log=False):
+    '''
+    Compute the transportation matrix to solve the regularized discrete
+        measures optimal transport max problem
+
+    The function solves the following optimization problem:
+    .. math::
```
rflamary: same here

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+##############################################################################
+
+
+def batch_grad_dual_alpha(M, reg, alpha, beta, batch_size, batch_alpha,
+                          batch_beta):
+    '''
+    Computes the partial gradient of F_\W_varepsilon
+
+    Compute the partial gradient of the dual problem:
+    ..Math:
```
rflamary: same here, use ```math``` instead of ```Math```

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+    return grad_alpha
+
+
+def batch_grad_dual_beta(M, reg, alpha, beta, batch_size, batch_alpha,
+                         batch_beta):
+    '''
+    Computes the partial gradient of F_\W_varepsilon
+
+    Compute the partial gradient of the dual problem:
+    ..Math:
```
rflamary: same here, use math instead of Math

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+
+
+def sgd_entropic_regularization(M, reg, batch_size, numItermax, lr,
+                                alternate=True):
+    '''
+    Compute the sgd algorithm to solve the regularized discrete measures
+        optimal transport dual problem
+
+    The function solves the following optimization problem:
+    .. math::
```
rflamary: same here

ot/stochastic.py
```diff
@@ -0,0 +1,791 @@
...
+
+
+def solve_dual_entropic(a, b, M, reg, batch_size, numItermax=10000, lr=1,
+                        log=False):
+    '''
+    Compute the transportation matrix to solve the regularized discrete measures
+        optimal transport dual problem
+
+    The function solves the following optimization problem:
+    .. math::
```
rflamary: same here, use math instead of Math

kilianFatras:
I think that SGD's performance also depends on the hyperparameters. It is really hard to find the good batch size or the learning rate. I tried many but as it is the dual problem, it is not as easy as for the semi dual problem. In [Genevay & al.], they pointed out that for the semi dual problem, you can have a lower bound of the Lipschitz constant which is not the case for the dual problem. I will keep working on this as it can increase the convergence speed.

PR #56, opened 2018-07-05
Free support barycenters

vivienseguy:
2-Wasserstein Barycenter algorithm with an example script. Optimization is carried over the locations of the support (not the weights). Only the unregularized case now. Regularized version will come soon.

### DIFF
--- a/examples/plot_free_support_barycenter.py
+++ b/examples/plot_free_support_barycenter.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """
 ====================================================
-2D Wasserstein barycenters of distributions
+2D free support Wasserstein barycenters of distributions
 ====================================================
 
 Illustration of 2D Wasserstein barycenters if discributions that are weighted
@@ -15,7 +15,8 @@ sum of diracs.
 
 import numpy as np
 import matplotlib.pylab as pl
-import ot.plot
+import ot
+
 
 ##############################################################################
 # Generate data
@@ -28,16 +29,16 @@ measures_weights = []
 
 for i in range(N):
 
-    n = np.random.randint(low=1, high=20)  # nb samples
+    n_i = np.random.randint(low=1, high=20)  # nb samples
 
-    mu = np.random.normal(0., 4., (d,))
+    mu_i = np.random.normal(0., 4., (d,))  # Gaussian mean
 
-    A = np.random.rand(d, d)
-    cov = np.dot(A, A.transpose())
+    A_i = np.random.rand(d, d)
+    cov_i = np.dot(A_i, A_i.transpose())  # Gaussian covariance matrix
 
-    x_i = ot.datasets.make_2D_samples_gauss(n, mu, cov)
-    b_i = np.random.uniform(0., 1., (n,))
-    b_i = b_i / np.sum(b_i)
+    x_i = ot.datasets.make_2D_samples_gauss(n_i, mu_i, cov_i)  # Dirac locations
+    b_i = np.random.uniform(0., 1., (n_i,))
+    b_i = b_i / np.sum(b_i)  # Dirac weights
 
     measures_locations.append(x_i)
     measures_weights.append(b_i)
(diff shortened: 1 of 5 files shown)


examples/plot_free_support_barycenter.py
```diff
@@ -0,0 +1,70 @@
...
+    b_i = np.random.uniform(0., 1., (n,))
+    b_i = b_i / np.sum(b_i)
+
+    measures_locations.append(x_i)
+    measures_weights.append(b_i)
+
+
+##############################################################################
+# Compute free support barycenter
+# -------------
```
rflamary: ```----``` needs to have the proper length for good documentation generation.

ot/lp/cvx.py
```diff
@@ -26,7 +27,7 @@ def scipy_sparse_to_spmatrix(A):
 
 
 def barycenter(A, M, weights=None, verbose=False, log=False, solver='interior-point'):
-    """Compute the entropic regularized wasserstein barycenter of distributions A
+    """Compute the Wasserstein barycenter of distributions A
```
rflamary: good catch !

ot/lp/cvx.py
```diff
@@ -144,3 +145,83 @@ def barycenter(A, M, weights=None, verbose=False, log=False, solver='interior-po
...
+    Solves the free support (locations of the barycenters are optimized, not the weights) Wasserstein barycenter problem (i.e. the weighted Frechet mean for the 2-Wasserstein distance)
+
+    The function solves the Wasserstein barycenter problem when the barycenter measure is constrained to be supported on k atoms.
+    This problem is considered in [1] (Algorithm 2). There are two differences with the following codes:
+    - we do not optimize over the weights
+    - we do not do line search for the locations updates, we use i.e. theta = 1 in [1] (Algorithm 2). This can be seen as a discrete implementation of the fixed-point algorithm of [2] proposed in the continuous setting.
+
+    Parameters
+    ----------
+    data_positions : list of (k_i,d) np.ndarray
```
rflamary: names in the documentation different from the code : data_positions vs measures_locations

ot/lp/cvx.py
```diff
@@ -144,3 +145,83 @@ def barycenter(A, M, weights=None, verbose=False, log=False, solver='interior-po
...
+    weights : (k,) np.ndarray
+        Initialization of the coefficients of the barycenter (non-negatives, sum to 1)
+
+    numItermax : int, optional
+        Max number of iterations
+    stopThr : float, optional
+        Stop threshol on error (>0)
+    verbose : bool, optional
+        Print information along iterations
+    log : bool, optional
```
rflamary: missing log parameter in the function.

would be nice to return the list of the displacement_square_norm along the iteration in a dictionnary if log=True (similar behavior as barycenter function above that retruns a log)

ot/lp/cvx.py
```diff
@@ -144,3 +145,83 @@ def barycenter(A, M, weights=None, verbose=False, log=False, solver='interior-po
...
+
+        displacement_square_norm = np.sum(np.square(X - T_sum))
+        X = T_sum
+
+        if verbose:
+            print('iteration %d, displacement_square_norm=%f\n', iter_count, displacement_square_norm)
+
+        iter_count += 1
+
+    return X
```
rflamary: add log if log=True

ot/lp/cvx.py
```diff
@@ -10,6 +10,7 @@
 import numpy as np
 import scipy as sp
 import scipy.sparse as sps
+import ot
```
rflamary: you shouldn't import pot inside a module.

something with relative path like
```python
from .__init__ import emd
```
is far better since it imports the emd function from ```the __init__.py```

ot/lp/cvx.py
```diff
@@ -144,3 +145,83 @@ def barycenter(A, M, weights=None, verbose=False, log=False, solver='interior-po
         return b, sol
     else:
         return b
+
+
+def free_support_barycenter(measures_locations, measures_weights, X_init, b, weights=None, numItermax=100, stopThr=1e-6, verbose=False):
```
rflamary: also do b=None if the weights are supposed uniform (needs  test an initialization in the function)

examples/plot_free_support_barycenter.py
```diff
@@ -0,0 +1,70 @@
...
+
+k = 10
+X_init = np.random.normal(0., 1., (k, d))
+b = np.ones((k,)) / k
+
+X = ot.lp.cvx.free_support_barycenter(measures_locations, measures_weights, X_init, b)
```
rflamary: ``` ot.lp.cvx.free_support_barycenter``` is very long.

you should import the function in ot.lp ```__init__.py``` and add it to ```__all__``` like barycenter so that you can do ``` ot.lp.free_support_barycenter```

PR #66, opened 2018-09-24
[MRG] Adding greenkhorn

arakotom:
* added greenkhorn algorithm in bregman.py
* added novel option for resolution
* modified unit test in test_variants


ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+        log dictionary return only if log==True in parameters
+
+    Examples
+    --------
+
+    >>> import ot
+    >>> a=[.5,.5]
+    >>> b=[.5,.5]
+    >>> M=[[0.,1.],[1.,0.]]
+    >>> ot.sinkhorn(a,b,M,1)
```
agramfort: ot.greenkhorn

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+
+    """
+    
+    i = 0
+    
+    n = a.shape[0]
+    m = b.shape[0]
+    
+    # Next 3 lines equivalent to K= np.exp(-M/reg), but faster to compute
+    K = np.empty(M.shape, dtype=M.dtype)
```
agramfort: K = np.empty_like(M)

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+    np.divide(M, -reg, out=K)
+    np.exp(K, out=K)
+    
+    u = np.ones(n)/n
```
agramfort: np.full(n, 1. / n)

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+    v = np.ones(m)/m
```
agramfort: np.full(m, 1. / m)

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+    G = np.diag(u)@K@np.diag(v)
```
agramfort: use broadcasting to avoid filling diagonal matrices

`G = u[:, np.newaxis] * K * v[, np.newaxis]`

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+            viol[i_1] = u[i_1]*K[i_1,:]@v - a[i_1]
+            viol_2 = viol_2 + ( K[i_1,:].T*(u[i_1] - old_u)*v) 
+
+        else:
+            old_v = v[i_2]
+            v[i_2] = b[i_2]/(K[:,i_2].T@u)
+            G[:,i_2] = u*K[:,i_2]*v[i_2]
+            #aviol = (G@one_m - a)
+            #aviol_2 = (G.T@one_n - b)
+            viol = viol + ( -old_v + v[i_2])*K[:,i_2]*u
```
agramfort: viol += ...

test/test_bregman.py
```diff
@@ -71,12 +71,14 @@ def test_sinkhorn_variants():
...
         u, u, M, 1, method='sinkhorn_epsilon_scaling', stopThr=1e-10)
     Gerr = ot.sinkhorn(u, u, M, 1, method='do_not_exists', stopThr=1e-10)
+    G_green = ot.sinkhorn(u, u, M, 1, method='greenkhorn', stopThr=1e-10)
 
     # check values
     np.testing.assert_allclose(G0, Gs, atol=1e-05)
     np.testing.assert_allclose(G0, Ges, atol=1e-05)
     np.testing.assert_allclose(G0, Gerr)
-
+    np.testing.assert_allclose(G0, G_green, atol =  1e-32)
```
agramfort: a pep8 checker would tell you but you should not put spaces around = in function signatures. It's to visually distinguish what is a function parameter from a variable assignment.

test/test_bregman.py
```diff
@@ -71,12 +71,14 @@ def test_sinkhorn_variants():
...
+    print(G0,G_green)
```
agramfort: and you should always put a space after a ,

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+    np.exp(K, out=K)
+    
+    u = np.ones(n)/n
+    v = np.ones(m)/m
+    G = np.diag(u)@K@np.diag(v)
+    
+    one_n = np.ones(n)
+    one_m = np.ones(m)
+    viol = G@one_m - a
+    viol_2 = G.T@one_n - b
```
agramfort: here to allocate arrays of ones to compute sum of rows and columns. I would just use np.sum(..., axis=)

ot/bregman.py
```diff
@@ -410,6 +417,148 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+    stopThr_val = 1
+    if log:
+        log['u'] = u
+        log['v'] = v
+
+    while i < numItermax and stopThr_val > stopThr:
```
agramfort: rather than using a while you could use a for loop. For optim solvers I tend to do:

```
for i in range(numItermax):
      ...
      if stopping condition satisfied do:
              break
else:
     print("Solver did not converge")
```

so you can easily print a message when you did not converge.

ot/bregman.py
```diff
@@ -410,6 +417,150 @@ def sinkhorn_knopp(a, b, M, reg, numItermax=1000,
...
+
+
+    See Also
+    --------
+    ot.lp.emd : Unregularized OT
+    ot.optim.cg : General regularized OT
+
+    """
+
+    i = 0
```
agramfort: not needed

agramfort: not needed

rflamary: indeed, thougt I got rid of all of them...

Issue #71, opened 2018-10-19
Compilation issue with MacOSX Mojave


ncourty:
I encountered the following issue while installing POT on MacOSX Mojave, with 
python 3.6 

`python setup.py build
running build
running build_py
running build_ext
building 'ot.lp.emd_wrap' extension
/usr/bin/gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/anaconda3/include -arch x86_64 -I/anaconda3/include -arch x86_64 -Iot/lp -I/anaconda3/lib/python3.6/site-packages/numpy/core/include -I/Users/nico/code/POT/ot/lp -I/anaconda3/include/python3.6m -c ot/lp/emd_wrap.cpp -o build/temp.macosx-10.7-x86_64-3.6/ot/lp/emd_wrap.o
warning: include path for stdlibc++ headers not found; pass '-std=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
In file included from ot/lp/emd_wrap.cpp:648:
In file included from /anaconda3/lib/python3.6/site-packages/numpy/core/include/numpy/arrayobject.h:4:
In file included from /anaconda3/lib/python3.6/site-packages/numpy/core/include/numpy/ndarrayobject.h:18:
In file included from /anaconda3/lib/python3.6/site-packages/numpy/core/include/numpy/ndarraytypes.h:1823:
/anaconda3/lib/python3.6/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:15:2: warning: "Using deprecated NumPy API, disable it by "
      "#defining NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-W#warnings]
#warning "Using deprecated NumPy API, disable it by " \
 ^
In file included from ot/lp/emd_wrap.cpp:650:
ot/lp/EMD.h:19:10: fatal error: 'iostream' file not found
#include <iostream>
         ^~~~~~~~~~
2 warnings and 1 error generated.
error: command '/usr/bin/gcc' failed with exit status 1`

I finally solved it by adding in setup.py the following extra argument for the compiler 
extra_compile_args=["-stdlib=libc++"]

However before pushing a PR, it is not clear to me if adding this option will break compatibility with other OS. Meanwhile, it is a simple workaround for this problem.

rflamary:
Hello @ncourty 

[...]

PR #72, opened 2018-10-19
add -tdlib=libc++ argument

ncourty:


rflamary:
OK so now we know this parameter breaks the linux version ;)

So now we need to have a test about if Macosx and if Mojave (or above?) in the setup.py I guess.

You ok with this @ncourty ?

ncourty:
Added a platform dependant test in order to solve permanently #71

setup.py
```diff
@@ -24,6 +24,12 @@
 with open(os.path.join(ROOT, 'README.md'), encoding="utf-8") as f:
     README = f.read()
 
+# add platform dependant optional compilation argument
+opt_arg=["-O3"]
+import platform
+if platform.system()=='Darwin':
+  if platform.release()=='18.0.0':
```
rflamary: shouldn't you use >= instead, i don't think this option will be necessary on just of release of MacSOX

```suggestion
  if platform.release()>='18.0.0':
```

PS trying the suggestion thing it's nice.

setup.py
```diff
@@ -24,6 +24,12 @@
 with open(os.path.join(ROOT, 'README.md'), encoding="utf-8") as f:
     README = f.read()
 
+# add platform dependant optional compilation argument
+opt_arg=["-O3"]
```
rflamary: good catch

setup.py
ncourty: not sure about the suggestion. In its form I think the bug is partly due to the fact that python Extension module has trouble to get the correct path from the new locations of clang compiler on the new OSX. I hope this will be corrected in the future. Clearly it is a temporary patch here.

PR #80, opened 2019-03-29
add empirical sinkhorn and sinkhorn divergence functions

kilianFatras:
Hello,

I am sending a PR. The PR has for purpose to add empirical functions. Mainly, the added functions just need the original source data, target data and the regularization parameter for entropic OT. In the PR, you will find:
- Function to calculate the OT transportation matrix from empirical data
- Function to calculate the OT loss from empirical data
- Function to calculate the Sinkhorn divergence from empirical data
- Tests for those functions
- Example
- Fix doc stochastic OT

Those functions will be in the bregman.py file. Their test functions will be in test_bregman.py. Finally, their examples have been put in plot_OT_2D_samples.py.

### DIFF
--- a/ot/bregman.py
+++ b/ot/bregman.py
@@ -1569,8 +1569,26 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
 
     .. [23] Aude Genevay, Gabriel Peyré, Marco Cuturi, Learning Generative Models with Sinkhorn Divergences,  Proceedings of the Twenty-First International Conference on Artficial Intelligence and Statistics, (AISTATS) 21, 2018
     '''
+    if log:
+        sinkhorn_loss_ab, log_ab = empirical_sinkhorn2(X_s, X_t, reg, a, b, metric=metric, numIterMax=numIterMax, stopThr=1e-9, verbose=verbose, log=log, **kwargs)
 
-    sinkhorn_div = (2 * empirical_sinkhorn2(X_s, X_t, reg, a, b, metric=metric, numIterMax=numIterMax, stopThr=1e-9, verbose=verbose, log=log, **kwargs) -
-                    empirical_sinkhorn2(X_s, X_s, reg, a, b, metric=metric, numIterMax=numIterMax, stopThr=1e-9, verbose=verbose, log=log, **kwargs) -
-                    empirical_sinkhorn2(X_t, X_t, reg, a, b, metric=metric, numIterMax=numIterMax, stopThr=1e-9, verbose=verbose, log=log, **kwargs))
-    return max(0, sinkhorn_div)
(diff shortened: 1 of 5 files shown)


rflamary:
Hello @kilianFatras and thank you for the PR,

Could you please build the html documentation (folder /docs, execute make html) and check that all the doc for the new function compile OK?

kilianFatras:
Hello, thank you for your answer.

I updated ot.bregman file with a new doc and I also updated ot.stochastic with a new doc.

ot/bregman.py
```diff
@@ -1296,3 +1298,279 @@ def unmix(a, D, M, M0, h0, reg, reg0, alpha, numItermax=1000,
...
+    >>> [2.99977435]
+
+
+    References
+    ----------
+
+    .. [23] Aude Genevay, Gabriel Peyré, Marco Cuturi, Learning Generative Models with Sinkhorn Divergences,  Proceedings of the Twenty-First International Conference on Artficial Intelligence and Statistics, (AISTATS) 21, 2018
+    '''
+
+    sinkhorn_div = (2 * empirical_sinkhorn2(X_s, X_t, reg, a, b, metric=metric, numIterMax=numIterMax, stopThr=1e-9, verbose=verbose, log=log, **kwargs) -
```
rflamary: This function fails if `log=True`. as a matter of fact if log=true you should return a log containing all 3 logs + the loss for each of the 3 OT compitation.

We also need a test for log=True

test/test_bregman.py
```diff
@@ -187,3 +188,57 @@ def test_unmix():
...
+    a = ot.unif(n)
+    b = ot.unif(n)
+    X_s = np.reshape(np.arange(n), (n, 1))
+    X_t = np.reshape(np.arange(0, n * 2, 2), (n, 1))
+    M = ot.dist(X_s, X_t)
+    M_s = ot.dist(X_s, X_s)
+    M_t = ot.dist(X_t, X_t)
+
+    emp_sinkhorn_div = ot.bregman.empirical_sinkhorn_divergence(X_s, X_t, 1)
+    sinkhorn_div = (2 * ot.sinkhorn2(a, b, M, 1) - ot.sinkhorn2(a, a, M_s, 1) -
```
rflamary: Definitely add a test for log=True here

kilianFatras: I added 2 log tests, I also changed the sinkhorn div = ot.sinkhorn2(a, b, M, 1) - 1/2 * ot.sinkhorn2(a, a, M_s, 1) - 1/2 * ot.sinkhorn2(b, b, M_t, 1)

davibarreira:
It seems that what `pot.py` is calling `empircal_sinkhorn_divergence` is different from the definition present in the paper by Genevay, which is cited in the documentation. What the paper calls "Sinkhorn Divergece" does not compute the regularization terms in the final cost, only for finding the optimal plan. Am I missing something? Is returning the "loss" instead of the cost using the entropical optimal plan done on purpose?

PR #91, opened 2019-07-05
[MRG] Release 0.6

rflamary:
This PR is a preparation for release 0.6 of POT

After all this work (and use of POT by me and other people) it feels as we reached a stable and nice list of tools around OT.

What I will do it this PR:
- [x] Change version to 0.6
- [x] Add detailed release text in RELEASES.md
- [x] Add build of wheels for manylinux using cibuildwheel (not on travis but locally before uploading to PIP)
- [x] Remove beta flag
- [x] Update Pypi classifiers
- [x] Add Python 3.7  to travis and remove 3.4


hichamjanati:
Hey @rflamary 
Indeed POT is used by many people -- including me -- but in my humble opinion, a 1.0 release should be reserved for "major incompatible and necessary changes planned long in advanced" (https://devguide.python.org/devcycle/)
(scikit-learn is still at 0.21 for instance). 

Several improvements can be made before preparing a 1.0 release. Off the top of my head:

* large sections of code are duplicated in some functions of the `bregman` module
* the naming of the funcs `sinkhorn` and `sinkhorn2` returning either the distance or the transport plan should be changed
* In some funcs, the transport plan computed (which can be heavy to store on gpus) even though it is not needed. I'm thinking there should be a function that explicitly computes the transport plan given the dual variables making the call specific.
* `sinkhorn` returns the distance or the plan depending on the second dimension of the input distribution `b` .. 
* The `emd` functions should be in a specific module not in the `__init__` file
* There is lot of commented code around
* the naming convention can be clearer and more consistent
* a two letters package name  -- `ot` --  can cause multiple headaches .. 


I know some of these are major changes but I think it would be better to discuss them now (it will only get harder to address them as the codebase grows larger) and continue to make minor / micro releases with deprecation warnings while doing a general review of the code, creating issues and working on them (potentially in a coding sprint :) ?), and then maybe prepare the 1.0 release.

hichamjanati:
I guess I'll finish the 3rd check in another PR then :)

ngayraud:
I completely agree with @hichamjanati . Especially on the naming issues: there are so many variables with names like a, b, etc... which should change into meaningful variable names - including the filenames (eg da could be named DomainAdaptation) A lot of little issues here and there need to be addressed. For example, in the da code, the fit() functions take Xs, ys, Xt, yt as input. Very often, one does not have (or need) ys and yt. The resulting code looks like this:
...fit(Xs, None, Xt). The order should change into (Xs, Xt, ys, yt).

A coding spring would surely address all of that, I'd be happy to attend if I can!

rflamary:
Hello,

I'm OK to call it release 0.6 but we need a proper release shortly (we are beginning to be very late on pip and conda) and a lot of the new stuff is only available on source. My idea behind 1.0 was that similarly to recent scipy i think we have achieved an important milestone and it might be the time to release a stable version that has been used in the wild. we could then start work on 2.0 that will introduce some API changes.

I also agree that POT could do with some work on the API and a cleanup of parameter names. But those would come with a proper deprecation schedule and I proper discussions between all contributors (and users).  

@hichamjanati I agree with most of your wish list but i will not put some API changes in the next release. The last release is already quite old and I will do a merge/release in the next weeks. I think the change you propose require some time to discuss and test before release because it will break a lot of things.

Meeting together for a code sprint is a very nice idea. I will think about how to organize this (we can also imagine a virtual meeting to limit transport).

rflamary:
Hello I have built wheels for all linux version (Py 2.7,3.4,3.5,3.6,3.7 for 32 and 64 bits). I will upload them on release so that it will not be necessary to compile POT from source when installing from Pypi.

We have a nice version 0.6 with the last unbalanced merge. The last release on pypi was nearly one year ago so I will go ahead with merging and releasing shortly this new version of POT if there is no objections.

I have seen a lot of comments (and upvotes) about API updates and I'm open to discussions. It's something that we can do with proper deprecation releases and so on but we need to discuss it all together. Maybe we should open a thread on the POT slack ?

ncourty:
> Le 9 sept. 2019 à 16:00, Rémi Flamary <notifications@github.com> a écrit :
> 
> Hello I have built wheels for all linux version (Py 2.7,3.4,3.5,3.6,3.7 for 32 and 64 bits). I will upload them on release so that it will not be necessary to compile POT from source when installing from Pypi.
> 
> We have a nice version 0.6 with the last unbalanced merge. The last release on pypi was nearly one year ago so I will go ahead with merging and releasing shortly this new version of POT if there is no objections.
> 
> I have seen a lot of comments (and upvotes) about API updates and I'm open to discussions. It's something that we can do with proper deprecation releases and so on but we need to discuss it all together. Maybe we should open a thread on the POT slack ?
> 

Yes that definitely a good idea !

Nicolas

agramfort:
ok

>

PR #87, opened 2019-06-12
[MRG] Add Unbalanced KL Wasserstein distance + barycenter

hichamjanati:
new unbalanced module for UOT with KL relaxation with the funcs:
- `sinkhorn_unbalanced`:  generalized Sinkhorn to compute W
- `barycenter_unbalanced`: unbalanced Wasserstein barycenter

- Tests of convergence for both algorithms
- Examples `plot_UOT_1D` and `plot_UOT_barycenter_1D` with unbalanced gaussian distributions.

PR #95, opened 2019-07-22
[WIP] Add stabilization for the unbalanced module

hichamjanati:
This PR complements #87 by adding:

- [x] replace the square by a max in the stopping criterion because the scaling blow up. This should also be done in `bregman.py`
- [x] `sinkhorn_stabilized_unbalanced` to compute the unbalanced optimal cost / transport plan with log stabilization. 
- [ ] `barycenter_unbalanced_log` to compute the unbalanced barycenter in log-domain with logsumexp operations. 

I am also thinking when a numerical error is encountered, Instead of stopping and returning the previous value, we could automatically switch to log-domain, what do you think ?

### DIFF
--- a/ot/unbalanced.py
+++ b/ot/unbalanced.py
@@ -77,8 +77,8 @@ def sinkhorn_unbalanced(a, b, M, reg, mu, method='sinkhorn', numItermax=1000,
     >>> b=[.5, .5]
     >>> M=[[0., 1.], [1., 0.]]
     >>> ot.sinkhorn_unbalanced(a, b, M, 1, 1)
-    array([[0.51122823, 0.18807035],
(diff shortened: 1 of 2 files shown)


hichamjanati:
I guess I'll finish the 3rd check in another PR then :)

rflamary:
Sorry mistake on my part i guess, I wanted to merge PR #97, but this one became merged?

I really only merge the other I dont know what happened.

Please do another PR indeed ;)

rflamary:
OK this is because you merged your other PR in #97.

Next time let me merge the small stuff  into master before merging between you PR and this will not happen.

hichamjanati:
Agh right, my bad. 
I'll make another PR for log-domain barycenters :)

PR #100, opened 2019-08-12
[MRG] Adds Unbalaced transport to domain adaptation methods + bugfixes

ngayraud:
This PR is primarily adds the unbalanced sinkhorn method to the domain adaptation classes. When the other sinkhorn methods will be implemented (eg stabilized) they will also be supported by the parameter method, which will now raise an error according to what is in unbalanced.py

In addition: 

- a small bug in the unbalanced.py warnings is corrected.
- Giving a norm value other than what is allowed and implemented will raise an error. I did that because I accidentally wrote something else and could not figure out where the bug was. Can be modified to raise a warning, but I think it should be an error.

### DIFF
--- a/ot/da.py
+++ b/ot/da.py
@@ -6,6 +6,7 @@ Domain adaptation with optimal transport
 # Author: Remi Flamary <remi.flamary@unice.fr>
 #         Nicolas Courty <ncourty@irisa.fr>
 #         Michael Perrot <michael.perrot@univ-st-etienne.fr>
+#         Nathalie Gayraud <nat.gayraud@gmail.com>
 #
 # License: MIT License
 
@@ -16,6 +17,7 @@ from .bregman import sinkhorn
 from .lp import emd
 from .utils import unif, dist, kernel, cost_normalization
 from .utils import check_params, BaseEstimator
+from .unbalanced import sinkhorn_unbalanced
 from .optim import cg
 from .optim import gcg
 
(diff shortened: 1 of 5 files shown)


hichamjanati:
Check these commits on my ongoing PR to fix travis
https://github.com/rflamary/POT/pull/99/commits/2566e976e66c47d28ac14bf4b6adc4a556159000 
https://github.com/rflamary/POT/pull/99/commits/8a9756dd0ce7c9fc0f7e8cf68c87654b65d5f31d

ot/da.py
```diff
@@ -1793,3 +1795,122 @@ def transform(self, Xs):
...
             return transp_Xs
+
+
+class UnbalancedSinkhornTransport(BaseTransport):
+
+    """Domain Adapatation unbalanced OT method based on sinkhorn algorithm
+
+    Parameters
+    ----------
+    reg_e : float, optional (default=1)
```
hichamjanati: Perhaps the default `reg_e` should be smaller ? or None and set to 10 / dimension for e.g in `fit`

ot/da.py
```diff
@@ -1793,3 +1795,122 @@ def transform(self, Xs):
...
+    ----------
+
+    .. [1] Chizat, L., Peyré, G., Schmitzer, B., & Vialard, F. X. (2016).
+    Scaling algorithms for unbalanced transport problems. arXiv preprint
+    arXiv:1607.05816.
+
+    """
+
+    def __init__(self, reg_e=1., reg_m=0.1, method='sinkhorn',
+                 max_iter=10, tol=10e-9, verbose=False, log=False,
```
hichamjanati: this `tol` is 1e-8 ?

ngayraud: oh yes. Actually the default value is 1e-9. Will fix.

ot/da.py
ngayraud: I agree, but since by default norm = None, maybe we could leave it like that to avoid numerical errors in the default run?

ngayraud: Actually I just realized that I have to check if the norm is None and do no normalization in that case, otherwise the error message will be triggered...

ngayraud: Ok it is done and it seems ok

ngayraud:
Shoot, there are still some syntax error and my flake8 and pydocstyle are buggy, I have to find a way to fix them

ot/da.py
hichamjanati: Ok I just realized the other `da` classes use the same default

ngayraud:
Aaah finally. We are green.

rflamary:
Thank you @ngayraud,

This PR looks very nice. I will have a quick look in the next days and will do the merge.

PR #106, opened 2019-10-19
MRG: Forgotten weights arg in barycenter funcs

hichamjanati:
- [x] Tiny fix: the `weights` arg was not passed in the general barycenter funcs.
- [x]  Loop counter in `barycenter_unbalanced` was outside the loop -> change `while` loops to `for` loops
- [x]  Adding 1e-16 to avoid 0 when dual variables are returned was too conservative and causes large differences in the exponentials. changed to 1e-300


rflamary:
Thank you Hicham !

PR #109, opened 2019-12-02
[MRG] Sparse emd solution

rflamary:
In this PR, i work on making the emd function able to return sparse OT matrices.

### DIFF
--- a/ot/lp/__init__.py
+++ b/ot/lp/__init__.py
@@ -27,7 +27,7 @@ __all__=['emd', 'emd2', 'barycenter', 'free_support_barycenter', 'cvx',
          'emd_1d', 'emd2_1d', 'wasserstein_1d']
 
 
-def emd(a, b, M, numItermax=100000, log=False):
+def emd(a, b, M, numItermax=100000, log=False, sparse=False):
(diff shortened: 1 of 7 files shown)


ot/lp/__init__.py
```diff
@@ -103,13 +107,20 @@ def emd(a, b, M, numItermax=100000, log=False):
     b = np.asarray(b, dtype=np.float64)
     M = np.asarray(M, dtype=np.float64)
 
+    sparse= not dense
```
hichamjanati: Isn't this redundant ? `if not dense` .. `else` ?

test/test_ot.py
```diff
@@ -118,6 +118,28 @@ def test_emd_empty():
...
+
+    n = 100
+    rng = np.random.RandomState(0)
+
+    x = rng.randn(n, 2)
+    x2 = rng.randn(n, 2)
+
+    M = ot.dist(x, x2)
+
+    G = ot.emd([], [], M)
```
hichamjanati: I would explicitly pass `dense=True` for clarity (and in case that changes in the future)

PR #112, opened 2019-12-05
[WIP] New variable names

kilianFatras:
Hello,

I am sending this PR to correct the variable names. For now on, you will find

. ot/stochastic.py
. test/test_stochastic.py
. example/ plot_stochastic.py

which have been cleaned.

### DIFF
--- a/examples/plot_stochastic.py
+++ b/examples/plot_stochastic.py
@@ -33,7 +33,7 @@ import ot.plot
 n_source = 7
 n_target = 4
 reg = 1
-numItermax = 1000
+num_iter_max = 1000
 
 a = ot.utils.unif(n_source)
 b = ot.utils.unif(n_target)
@@ -53,7 +53,7 @@ M = ot.dist(X_source, Y_target)
 
 method = "SAG"
 sag_pi = ot.stochastic.solve_semi_dual_entropic(a, b, M, reg, method,
-                                                numItermax)
+                                                num_iter_max)
 print(sag_pi)
 
 #############################################################################
@@ -69,7 +69,7 @@ print(sag_pi)
 n_source = 7
 n_target = 4
 reg = 1
-numItermax = 1000
+num_iter_max = 1000
 log = True
 
 a = ot.utils.unif(n_source)
@@ -91,7 +91,7 @@ M = ot.dist(X_source, Y_target)
 
 method = "ASGD"
 asgd_pi, log_asgd = ot.stochastic.solve_semi_dual_entropic(a, b, M, reg, method,
-                                                           numItermax, log=log)
+                                                           num_iter_max, log=log)
 print(log_asgd['alpha'], log_asgd['beta'])
 print(asgd_pi)
 
@@ -153,7 +153,7 @@ pl.show()
 n_source = 7
 n_target = 4
 reg = 1
-numItermax = 100000
+num_iter_max = 100000
 lr = 0.1
 batch_size = 3
 log = True
@@ -175,7 +175,7 @@ M = ot.dist(X_source, Y_target)
 # Call ot.solve_dual_entropic and plot the results.
 
 sgd_dual_pi, log_sgd = ot.stochastic.solve_dual_entropic(a, b, M, reg,
-                                                         batch_size, numItermax,
+                                                         batch_size, num_iter_max,
                                                          lr, log=log)
 print(log_sgd['alpha'], log_sgd['beta'])
 print(sgd_dual_pi)
--- a/examples/plot_stochastic.py
+++ b/examples/plot_stochastic.py
@@ -33,7 +33,7 @@ import ot.plot
 n_source = 7
 n_target = 4
 reg = 1
-num_iter_max = 1000
+maxiter = 1000
 
 a = ot.utils.unif(n_source)
 b = ot.utils.unif(n_target)
@@ -53,7 +53,7 @@ M = ot.dist(X_source, Y_target)
 
 method = "SAG"
 sag_pi = ot.stochastic.solve_semi_dual_entropic(a, b, M, reg, method,
-                                                num_iter_max)
+                                                maxiter)
 print(sag_pi)
 
 #############################################################################
@@ -69,7 +69,7 @@ print(sag_pi)
 n_source = 7
 n_target = 4
 reg = 1
-num_iter_max = 1000
+maxiter = 1000
 log = True
 
 a = ot.utils.unif(n_source)
@@ -91,7 +91,7 @@ M = ot.dist(X_source, Y_target)
 
 method = "ASGD"
 asgd_pi, log_asgd = ot.stochastic.solve_semi_dual_entropic(a, b, M, reg, method,
-                                                           num_iter_max, log=log)
+                                                           maxiter, log=log)
 print(log_asgd['alpha'], log_asgd['beta'])
 print(asgd_pi)
 
@@ -153,7 +153,7 @@ pl.show()
 n_source = 7
 n_target = 4
 reg = 1
-num_iter_max = 100000
+maxiter = 100000
 lr = 0.1
 batch_size = 3
 log = True
@@ -175,7 +175,7 @@ M = ot.dist(X_source, Y_target)
 # Call ot.solve_dual_entropic and plot the results.
 
 sgd_dual_pi, log_sgd = ot.stochastic.solve_dual_entropic(a, b, M, reg,
-                                                         batch_size, num_iter_max,
+                                                         batch_size, maxiter,
                                                          lr, log=log)
 print(log_sgd['alpha'], log_sgd['beta'])
 print(sgd_dual_pi)
--- a/examples/plot_stochastic.py
+++ b/examples/plot_stochastic.py
@@ -33,7 +33,7 @@ import ot.plot
 n_source = 7
 n_target = 4
 reg = 1
-maxiter = 1000
+max_iter = 1000
 
 a = ot.utils.unif(n_source)
 b = ot.utils.unif(n_target)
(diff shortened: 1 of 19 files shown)


examples/plot_stochastic.py
```diff
@@ -33,7 +33,7 @@
 n_source = 7
 n_target = 4
 reg = 1
-numItermax = 1000
+num_iter_max = 1000
```
hichamjanati: Suggestion: `maxiter` would be more standard since it's used in scipy and sklearn ?

test/test_stochastic.py
```diff
@@ -60,18 +60,18 @@ def test_stochastic_sag():
...
-    n = 15
+    n_samples = 15
     reg = 1
-    numItermax = 100000
+    num_iter_max = 100000
     rng = np.random.RandomState(0)
 
-    x = rng.randn(n, 2)
-    u = ot.utils.unif(n)
+    x_samples = rng.randn(n_samples, 2)
```
hichamjanati: imho, I think `x` is fine or perhaps. `x_samples` and `n_samples` is perhaps too visually similar but denote very different objects.

kilianFatras: I agree that they are similar. I am fine with x, but I would keep n_samples to know what we are using here.

examples/plot_stochastic.py
kilianFatras: Are you talking about just the plot_stochastic module ? Or every function ?

ngayraud: I agree with @hichamjanati. Moreover, I think that if we change it here, we should change it everywhere.

ngayraud:
What about variables such as "a", "b" and "M"? Don't we want to make those more readable as well? Especially "M" which I understand is the cost?

examples/plot_stochastic.py
kilianFatras: I think if we change it somewhere, we should change it everywhere as well.

test/test_stochastic.py
```diff
@@ -90,19 +90,19 @@ def test_stochastic_asgd():
 
 def test_sag_asgd_sinkhorn():
     # test all algorithms
-    n = 15
+    n_samples = 15
     reg = 1
     nb_iter = 100000
```
ngayraud: why not rename this to num_iter_max as well? (or maxiter, depending on what we choose)

examples/plot_stochastic.py
agramfort: num_iter_max sounds really french. +1 for maxiter like sklearn

test/test_stochastic.py
```diff
@@ -90,19 +90,19 @@ def test_stochastic_asgd():
...
     rng = np.random.RandomState(0)
 
-    x = rng.randn(n, 2)
-    u = ot.utils.unif(n)
-    M = ot.dist(x, x)
+    x_samples = rng.randn(n_samples, 2)
```
ngayraud: You can maybe make it a capital X

test/test_stochastic.py
```diff
@@ -90,19 +90,19 @@ def test_stochastic_asgd():
...
+    u = ot.utils.unif(n_samples)
```
ngayraud: what is u?

examples/plot_stochastic.py
rflamary: I think maxiter is great  but Kilian this is not a one submodule change, you need to change it in the whole toolbox while mainteining compatibility with a deprecation warning.  I think it can be done with decorators where we give corresponding old_name-new name. In any case you should not change the tests (other than changing variable names), you should provide new test with the new parameters tested until we do a proper deprecation/removing of the old names.

Before rushing into things i think we should agree and discuss all variable change and how we can provide old compatibility without adding ugly ifs and having double variables in the function signature. I created a wiki page called Variable renaming where we can put all the propositions. unless there is already a document where this is discussed we really need to agree on things in order to avoid unecessary work.

rflamary:
Hello,

In order to keep compatibility you shoudl use this decorator (should be added in utils.py wit propoer documentation and exmaple)

```python
class deprecated_variable(object):

    def __init__(self, dic_change=None):
        self.dic_change = dic_change

    def __call__(self, fun):
        
        msg_template = "Function parameter '{}' is deprecated and should be replaced by '{}' " 

        def wrapped(*args, **kwargs):
            kwargs2=dict()
            for kw in kwargs:
                if kw in self.dic_change:
                    warnings.warn(msg_template.format(kw,self.dic_change[kw]), category=DeprecationWarning)
                    kwargs2[self.dic_change[kw]]=kwargs[kw]
                else:
                    kwargs2[kw]=kwargs[kw]
            return fun(*args, **kwargs2)

        wrapped.__name__ = fun.__name__
        wrapped.__dict__ = fun.__dict__
        wrapped.__doc__ = fun.__doc__
        
        return wrapped

```

on this example

```python
@deprecated_variable({'param2_old':'param2'})
def func(a,b,param1=0,param2=1):
    print(a,b,param1,param2)

```

you can call the function func with the new (param2) or old (param2_old) and it will work on both but print the following message if calling with the old parameter :

```
 DeprecationWarning: Function parameter 'param2_old' is deprecated and should be replaced by 'param2' 
```

The decorator can handle any number of parameter name  switches. it should provide the proper documentation for the function also.

Note that there is no need to do decoration if you change the name of non keyword variables since the code will still work.

kilianFatras:
Hello everyone,

I pushed a new version for variable and argument names. I included the decorator from @rflamary in utils.py and used it for ot.stochastic. You can see that it takes `numItermax` and pass to `maxiter`. Furthermore, I came back to old test (with numItermax as argument) and extended them with the new argument names + new variable name.

ngayraud:
The variable name for the maximum number of iterations is `max_iter` in sklearn and not `maxiter`, but I don't want to be too nitpicky. You can change it or not, your choice, I just thought it is goot to know since we used it to decide the name.

ot/stochastic.py
```diff
@@ -441,10 +444,10 @@ def solve_semi_dual_entropic(a, b, M, reg, method, numItermax=10000, lr=None,
           a[:, None] * b[None, :])
 
     if log:
-        log = {}
-        log['alpha'] = opt_alpha
-        log['beta'] = opt_beta
-        return pi, log
+        log_stoc = {}
+        log_stoc['alpha'] = opt_alpha
+        log_stoc['beta'] = opt_beta
```
ngayraud: `log_stoc = {'alpha': opt_alpha, 'beta': opt_beta}` ?

ot/stochastic.py
```diff
@@ -587,7 +591,7 @@ def sgd_entropic_regularization(a, b, M, reg, batch_size, numItermax, lr):
         Regularization term > 0
     batch_size : int
         size of the batch
-    numItermax : int
+    maxiter : int
         number of iteration
```
ngayraud: nit: iterations

ot/stochastic.py
```diff
@@ -743,13 +748,13 @@ def solve_dual_entropic(a, b, M, reg, batch_size, numItermax=10000, lr=1,
...
           a[:, None] * b[None, :])
     if log:
-        log = {}
-        log['alpha'] = opt_alpha
-        log['beta'] = opt_beta
-        return pi, log
+        log_stoc = {}
+        log_stoc['alpha'] = opt_alpha
+        log_stoc['beta'] = opt_beta
+        return pi, log_stoc
```
ngayraud: same as above!

ot/stochastic.py
hichamjanati: Are `alpha` and `beta` in log-domain ?

test/test_stochastic.py
hichamjanati: Yes agreed for `n_samples`. `X` is probably more appropriate for multivariate data.

kilianFatras:
@ngayraud I also prefer `max_iter` but I thought we were using `maxiter`. Anyway, this is the right time to decide which one we use. I also like the way to short `log_stoc`.

@hichamjanati gamma variable name comes from the documentation where we defined the OT plan as gamma. I would prefer to have a consistency between the documentation and the variable names.

ot/stochastic.py
kilianFatras: @hichamjanati What do you mean ? They are the usual dual variables.

hichamjanati: Ok so in log-domain (plan given by `exp(alpha / eps) K exp (beta / eps). 
(Some Sinkhorn funcs return `exp(alpha / eps)` instead of `alpha` which should be normalized to `alpha` everywhere)

ngayraud:
To be honest, `gamma` only "speaks" to people that are very familiar with the annotation. something like `ot_plan` would be more appropriate IMO. Also, if we can change `maxiter` to `max_iter`, it would b following the variable naming convention we are aiminng for.

ot/stochastic.py
kilianFatras: I agree, Sinkhorn functions return `exp(alpha/eps)` instead of alpha and it should be clarified in the documentation or normalized to `alpha`. I also prefer normalizing to `alpha`.

Regarding ot.stochastic, it returns the dual potentials `alpha`, `beta` (you can see when we compute the OT plan).

kilianFatras:
I agree with max_iter, I will update the PR. I am also fine with 'ot_plan', my point was that it needs to be consistent with the documentation and we will need to change it as well. @rflamary, @ncourty  what do you think ?

kilianFatras:
Hello all,

I have added new stuffs.
\gamma -> OT_{\text{plan}}
maxiter -> max_iter
correction of a bug for sphynx library.

rflamary:
Hello thank you all for all your work, we ware working on a new API for POT that will use more uniform variable name so i'm closing this PR.

PR #116, opened 2019-12-09
[MRG] Add assert for emd dimension mismatch

kilianFatras:
Hello,

I am sending a PR with a patch to issue 114. In this PR, you will find:

. An assert condition (M dimensions with a and b)
. Test to catch the Assertion error.


rflamary:
Hello @kilianFatras,

I'm ok with the PR but you need to resolve the conflict that appeared wrt the master branche before merge.

PR #121, opened 2020-01-06
[WIP] add screenkhorn: screening Sinkhorn algorithm

mzalaya:


ot/bregman.py
```diff
@@ -1787,3 +1789,287 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
...
+                aK_sort = np.sort(aK)[::-1]
+            epsilon_u_square = aK_sort[ns_budget - 1:ns_budget + 1].mean()
+            I = np.where(a >= epsilon_u_square * K_sum_cols)[0].tolist()
+
+        if len(J) != nt_budget:
+            if uniform:
+                bK = b / K_sum_rows
+                bK_sort = np.sort(bK)[::-1]
+            epsilon_v_square = bK_sort[ns_budget - 1:ns_budget + 1].mean()
+            J = np.where(b >= epsilon_v_square * K_sum_rows)[0].tolist()
```
rflamary: could you stay with array of integer?

i would be suprised if np.ix_ couldn't handle those.

ot/bregman.py
```diff
@@ -1787,3 +1789,287 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
...
+
+        if verbose:
+            print("Epsilon = %s\n" %epsilon)
+            print("Scaling factor = %s\n" %kappa)
+
+        if verbose:
+            print('|I_active| = %s \t |J_active| = %s ' %(len(I), len(J)))
+
+        # Ic, Jc: complementary of the active sets I and J
+        Ic = list(set(list(range(ns))) - set(I))
```
rflamary: numpy.delete on an numpy.arange should also work. you should test to see which one is the more efficient.

ot/bregman.py
```diff
@@ -1787,3 +1789,287 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
...
+            cpt += 1
+
+        u0 = projection(u0, epsilon / kappa)
+        v0 = projection(v0, epsilon * kappa)
+
+    else:
+        u0 = u0
+        v0 = v0
+
+    def restricted_sinkhorn(usc, vsc, max_iter=5):
```
hichamjanati: Naive question: why 5 for max_iter ?

ot/bregman.py
```diff
@@ -1787,3 +1789,287 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
...
+
+    usc = theta[:ns_budget]
+    vsc = theta[ns_budget:]
+
+    usc_full = np.full(ns, epsilon / kappa)
+    vsc_full = np.full(nt, epsilon * kappa)
+    usc_full[I] = usc
+    vsc_full[J] = vsc
+
+    Gsc = usc_full.reshape((-1, 1)) * K * vsc_full.reshape((1, -1))
```
hichamjanati: Pprefer the numpy slicing instead of reshape: 
```python
gamma = u[:, None] * K * v[None, :]
return gamma
```
This allows to "see" the formula gamma_ij = u_i K_ij v_j. 

This function should also have an argument `log` and return the transport and a dictionary `log` containing the dual variables (see the other sinkhorn functions).

ot/bregman.py
```diff
@@ -8,13 +8,15 @@
...
 #         Hicham Janati <hicham.janati@inria.fr>
+#         Mokhtar Z. Alaya <mokhtarzahdi.alaya@gmail.com>
 #
 # License: MIT License
 
 import numpy as np
 import warnings
 from .utils import unif, dist
-
+import bottleneck
```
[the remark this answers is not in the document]
mzalaya: I prefer to import bottleneck inside the function, I tested it and it works. I will mention this dependency in the documentation.

ot/bregman.py
```diff
@@ -1787,3 +1789,287 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
 
         sinkhorn_div = sinkhorn_loss_ab - 1 / 2 * (sinkhorn_loss_a + sinkhorn_loss_b)
         return max(0, sinkhorn_div)
+ 
+def screenkhorn(a, b, M, reg, ns_budget, nt_budget, uniform=True, restricted=True, verbose=False):
```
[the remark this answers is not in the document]
mzalaya: @hichamjanati: yes I agree with your proposition. I prefer using the test than the threshold. As ns_budget should be integer than the default value is ns_budget = int(np.floor(0.5*ns)) (the same for nt_budget).

ot/bregman.py
mzalaya: It's just a guess value :) to give a good warm-start point from the LBFGS solver.  I think it should be greater than 3 and less than 5 (the more given iterations in restricted_sinkhorn function, the more spending time for the LBFGS initialization)

ot/bregman.py
rflamary: agreed with both comments

ot/bregman.py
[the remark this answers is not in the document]
mzalaya: I think it's better to check from the beginning of the function if the bottleneck module is installed or not. I did that using python Exception.

ot/bregman.py
rflamary: binary indexing is better indeed and it is much easier do do the complementary.

mzalaya: Actually, I compared three ways to define Ic as follows:
```
fulllist_s = list(np.arange(ns))
Ic = list(np.delete(fulllist_s, [fulllist_s.index(v) for v in I]))
Ic = list(np.array(fulllist_s)[~np.isin(fulllist_s, I)])
Ic = list(set(fulllist_s) - set(I))
```
The definition using `set` is much faster than ways 1 and 2. So, I suggest to keep it.

hichamjanati: No, it's the conversion array to list to array that slows it down, if we're going for numpy, we must not use lists anywhere:
```python
fulllist_s = np.arange(ns)
Ic = fulllist_s[~np.isin(fulllist_s, I)]
```
 3x faster with numpy

Screenshot 2020-01-07 at 19 19 06

rflamary: yes binary indexing is faster but it needs to stay in full numpy. From the algorithm it seems to me it is feasible and the code might even be easier to understand with binary indexing.

rflamary:
Hello @mzalaya,

I see you have implemented a lot of our comments and the code is shaping up very well. 

There is still the binary indexing that needs to be looked at and I would also like you to add an example of you solver in the example folder of the repository. The examples are afterwards included semi automatically in the documentation with the plots and everything and it will definitely make your new contribution more visible. Also we need some tests on the new function that i don't see running in travis yet.

I triggered the build for your last commit to see if the tests pass. I don't know why it did not trigger itself (spoiler alert it failed because of pep8 violations).

rflamary:
Hello @mzalaya,

We are nearly there. What is still missing before merge:
- Test should pass: use autopep8 on the example file also
- Add a test that call your function ,and check that the output makes sens.
- you should use `r"""doc"""` for using the latex backslash in the documentatiion

mzalaya:
@rflamary  I think `screenkhorn` is ready to be merged in POT : )

ot/bregman.py
```diff
@@ -1787,3 +1789,339 @@ def empirical_sinkhorn_divergence(X_s, X_t, reg, a=None, b=None, metric='sqeucli
...
+    References
+    -----------
+    .. [26] Alaya M. Z., Bérar M., Gasso G., Rakotomamonjy A. (2019). Screening Sinkhorn Algorithm for Regularized Optimal Transport (NIPS) 33, 2019
+
+    """
+    # check if bottleneck module exists
+    try:
+        import bottleneck
+    except ImportError:
+        print("Bottleneck module doesn't exist. Install it from https://pypi.org/project/Bottleneck/")
```
rflamary: The function should not fail when bottleneck is not installed and it will because it uses bottleneck afterwards. As already said we don't want to add dependencies.

i would do something like 
```python
try:
 ￼   import bottleneck
except ImportError:
    warning("Bottleneck module is not installed. Install it from https://pypi.org/project/Bottleneck/ for better performance")
    bottleneck=np
```
this would use the numpy partition that is not as efficient but works on any numpy machine. I'm surprise the test passes.

mzalaya: good suggestion to sent a warning for non installed Bottelenck module.

rflamary:
Hello @mzalaya,

So now autograd breaks on python 2.7. While we could deactivate the tests on 2.7 (it is deprecated) is would like to force to use an older version of pymanopt for the tests.

for that i think we can change the line  `pymanopt` in requirement.txt and replace it with the two following lines:
```
pymanopt== 0.2.4, python_version < '3'
pymanopt, python_version >= '3'
```

Could you try to do that please? We will remove support for 2.7 at a future date but this is another PR ;)

mzalaya:
@rflamary  
a semicolon is needed  between. cool now it's working :) 
````pymanopt==0.2.4; python_version <'3'
       pymanopt; python_version >= '3'
````

PR #123, opened 2020-01-27
[WIP] Bug on dual potential (constraint violation when some weights are 0)

rflamary:
I just discovered this important bug in the `ot.emd``function and it feels like opening pandoras's box.

I added in the fisrt commit a test that checks that the constraints are satisfied and we can see in the travis build that the test fails. This comes from the fact that the c++ emd solver is sparse in the sens that if a weight is 0 it will discard the bin and solve a smaller subproblem on the histograms that have non-zero weights. It works very well and is very efficient in practice bug it does not return proper dual variables because it sets the value for the bins with 0 weight to 0 which can lead to constraint violation in practice.

What I will do in this PR:
- [x] Correct the bug:  `ot.emd` should work by default (return proper dual variables/subgradients)
- [x] add a postprocessing (as default but and ot.emd parameter) to the dual potential so that they are in a way 'centered' because the invariance to a constant can be a big problem in practice when using theme as subgradients. The pre-processing will not change the objective value but provide some stability (and will always return one of the possible solution).

The post-processing is actually necessary when trying to estimate correct dual variable from the optimal sub-problem if we want to avoid bias in the resulting dual potentials (bnasically everyone at 0 on either the left or right potential).

### DIFF
--- a/ot/lp/__init__.py
+++ b/ot/lp/__init__.py
@@ -28,10 +28,10 @@ __all__ = ['emd', 'emd2', 'barycenter', 'free_support_barycenter', 'cvx',
 
 
 def center_ot_dual(alpha0, beta0, a=None, b=None):
-    r"""Center dual OT potentials wrt theirs weights
+    r"""Center dual OT potentials w.r.t. theirs weights
 
     The main idea of this function is to find unique dual potentials
-    that ensure some kind of centering/fairness. It will help have
+    that ensure some kind of centering/fairness. The main idea is to find dual potentials that lead to the same final objective value for both source and targets (see below for more details). It will help having
     stability when multiple calling of the OT solver with small changes.
 
     Basically we add another constraint to the potential that will not
@@ -91,7 +91,15 @@ def center_ot_dual(alpha0, beta0, a=None, b=None):
 def estimate_dual_null_weights(alpha0, beta0, a, b, M):
     r"""Estimate feasible values for 0-weighted dual potentials
 
-    The feasible values are computed efficiently bjt rather coarsely.
+    The feasible values are computed efficiently but rather coarsely.
+
+    .. warning::
+        This function is necessary because the C++ solver in emd_c
+        discards all samples in the distributions with 
+        zeros weights. This means that while the primal variable (transport 
+        matrix) is exact, the solver only returns feasible dual potentials
+        on the samples with weights different from zero. 
+
     First we compute the constraints violations:
 
     .. math::
@@ -113,11 +121,11 @@ def estimate_dual_null_weights(alpha0, beta0, a, b, M):
 
         \beta_j = \beta_j -v^b_j \quad \text{ if } b_j=0 \text{ and } v^b_j>0
 
-    In the end the dual potential are centred using function
+    In the end the dual potentials are centered using function
     :ref:`center_ot_dual`.
 
     Note that all those updates do not change the objective value of the
-    solution but provide dual potential that do not violate the constraints.
+    solution but provide dual potentials that do not violate the constraints.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+    alpha = alpha0 + c
+    beta = beta0 - c
+
+    return alpha, beta
+
+
+def estimate_dual_null_weights(alpha0, beta0, a, b, M):
+    r"""Estimate feasible values for 0-weighted dual potentials
+
+    The feasible values are computed efficiently bjt rather coarsely.
```
ncourty: but

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+
+    Finally we update the dual potential with 0 weights if a
+    constraint is violated
+
+    .. math::
+        \alpha_i = \alpha_i -v^a_i \quad \text{ if } a_i=0 \text{ and } v^a_i>0
+
+        \beta_j = \beta_j -v^b_j \quad \text{ if } b_j=0 \text{ and } v^b_j>0
+
+    In the end the dual potential are centred using function
```
ncourty: potentials 
are centered

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+    :ref:`center_ot_dual`.
+
+    Note that all those updates do not change the objective value of the
+    solution but provide dual potential that do not violate the constraints.
```
ncourty: potentials

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+    alpha0 : (ns,) numpy.ndarray, float64
+        Source dual potential
+    beta0 : (nt,) numpy.ndarray, float64
+        Target dual potential
+    alpha0 : (ns,) numpy.ndarray, float64
+        Source dual potential
+    beta0 : (nt,) numpy.ndarray, float64
+        Target dual potential
+    a : (ns,) numpy.ndarray, float64
+        Source histogram (uniform weight if empty list)
```
ncourty: histogram -> distribution ? 
weight -> weights

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+    b : (nt,) numpy.ndarray, float64
+        Target histogram (uniform weight if empty list)
```
ncourty: same here

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+        Target corrected dual potential
+
+    """
+
+    # binary indexing of non-zeros weights
+    asel = a != 0
+    bsel = b != 0
+
+    # compute dual constraints violation
+    Viol = alpha0[:, None] + beta0[None, :] - M
```
ncourty: can we change the name of the variable here by something more explicit, e.g. constraints_violation ?

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+
+    # Compute worst violation per line and columns
```
ncourty: worst -> largest

test/test_ot.py
```diff
@@ -338,6 +338,10 @@ def test_dual_variables():
     np.testing.assert_almost_equal(cost1, log['cost'])
     check_duality_gap(a, b, M, G, log['u'], log['v'], log['cost'])
 
+    viol = log['u'][:, None] + log['v'][None, :] - M
```
ncourty: same here

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
 
-__all__=['emd', 'emd2', 'barycenter', 'free_support_barycenter', 'cvx',
-         'emd_1d', 'emd2_1d', 'wasserstein_1d']
+__all__ = ['emd', 'emd2', 'barycenter', 'free_support_barycenter', 'cvx',
+           'emd_1d', 'emd2_1d', 'wasserstein_1d']
 
 
-def emd(a, b, M, numItermax=100000, log=False, dense=True):
+def center_ot_dual(alpha0, beta0, a=None, b=None):
+    r"""Center dual OT potentials wrt theirs weights
```
ncourty: wrt -> wrt.

ot/lp/__init__.py
```diff
@@ -23,11 +23,150 @@
...
+
+    The main idea of this function is to find unique dual potentials
+    that ensure some kind of centering/fairness. It will help have
```
ncourty: can you be more explicit on the meaning of 'centering/fairness' ?

ncourty: have -> having

rflamary:
Thank you @ncourty i took care of your comments. We now have a warning in the emd_c function and the reason why we need to find proper dual potential is discussed.

PR #129, opened 2020-02-27
Gromov 1d

tvayer:
In this PR we propose to add the close form for Gromov-Wasserstein between 1d measures:
- Add functions that calculates the gromov for 1d measures in ot.gromov
- Add the test loops for these functions


rflamary:
Closing this PR that is now too old.

Issue #124, opened 2020-01-30
Bregman convolutional barycenter 2d return an error for non-square images


ThibautLG:
When ever I call 
`ot.bregman.convolutional_barycenter2d(A, reg, weights)`
with A whose shape is not of the form (k,n,n), the function returns an error.

In my case, using the same code as in the example in the [doc](https://pot.readthedocs.io/en/stable/auto_examples/plot_convolutional_barycenter.html), I get the following error for (600,500) images.

```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-f723cd8affb9> in <module>
     55         else:
     56             # call to barycenter computation
---> 57             pl.imshow(ot.bregman.convolutional_barycenter2d(A, reg, weights), cmap=cm)
     58             pl.axis('off')
     59 pl.show()

~/.local/lib/python3.6/site-packages/POT-0.6.0-py3.6-linux-x86_64.egg/ot/bregman.py in convolutional_barycenter2d(A, reg, weights, numItermax, stopThr, stabThr, verbose, log)
   1361         b = np.zeros_like(A[0, :, :])
   1362         for r in range(A.shape[0]):
-> 1363             KV[r, :, :] = K(A[r, :, :] / np.maximum(stabThr, K(U[r, :, :])))
   1364             b += weights[r] * np.log(np.maximum(stabThr, U[r, :, :] * KV[r, :, :]))
   1365         b = np.exp(b)

~/.local/lib/python3.6/site-packages/POT-0.6.0-py3.6-linux-x86_64.egg/ot/bregman.py in K(x)
   1352 
   1353     def K(x):
-> 1354         return np.dot(np.dot(xi1, x), xi1)
   1355 
   1356     while (err > stopThr and cpt < numItermax):

<__array_function__ internals> in dot(*args, **kwargs)

ValueError: shapes (600,500) and (600,600) not aligned: 500 (dim 1) != 600 (dim 0)
```
My system information:

> Linux-5.3.0-28-generic-x86_64-with-Ubuntu-18.04-bionic
> Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
> [GCC 8.3.0]
> NumPy 1.17.4
> SciPy 1.3.3
> POT 0.6.0

rflamary:
This issue has been clsed in PR #132

PR #132, opened 2020-03-05
[MRG] Fix convolutional_barycenter2d kernel for non-symmetric images

atong01:
Thank you for your awesome package!

The code and examples for `ot.bregman.convolutional_barycenter2d` assumes a symmetric image. This PR changes the kernel to allow for non-symmetric images and adds a simple test of this functionality.

This addresses Issue #124

### DIFF
--- a/ot/bregman.py
+++ b/ot/bregman.py
@@ -9,6 +9,7 @@ Bregman projections for regularized OT
 #         Titouan Vayer <titouan.vayer@irisa.fr>
 #         Hicham Janati <hicham.janati@inria.fr>
 #         Mokhtar Z. Alaya <mokhtarzahdi.alaya@gmail.com>
+#         Alexander Tong <alexander.tong@yale.edu>
 #
 # License: MIT License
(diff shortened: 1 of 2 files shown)


rflamary:
This is great, thank you for the PR.

I think one day we should implement a proper convolution instead of those matrix products (or even circular convolution with Fourier as an option). 

you can add you name at the top of the file as Author . I will do the merge in the next days. I am still waiting for a few things but I will do a new release shortly.

atong01:
Agreed a proper convolution does sound useful in some circumstances. Interesting to note a simple usage of `scipy.ndimage.filter.gaussian_filter` with this kernel is ~4X slower on the example images. I would guess matrix products are simply better optimized at this scale.
```
def K(x):
    # sigma to match previous behavior
    sigma = [np.sqrt(reg / 2) * dim for dim in x.shape]
    x = gaussian_filter(x, sigma=sigma, mode="constant")
    return x
```

PR #133, opened 2020-03-09
[MRG] Fix Fused Gromov Wasserstein implementation

little-nem:
The `alpha` trade-off parameter of the Fused Gromov Wasserstein implementation was not used correctly: what was computed was 
```
(arg)min_\gamma <\gamma, M>_F + \alpha \sum L(C1_{i,k},C2_{j,l}) T_{i,j} T_{k,l}
```
instead of the convex combination
```
(arg)min_\gamma (1 - \alpha) <\gamma, M>_F + \alpha \sum L(C1_{i,k},C2_{j,l}) T_{i,j} T_{k,l}
```
Even though the two are mathematically equivalent up to a rescaling of alpha, it was not consistent with the documentation. I changed the implementation so `optim.cg` is called with `cg(..., (1-alpha)*M, alpha, ...)` instead of `cg(..., M, alpha, ...)` which should be enough to fix this issue.

### DIFF
--- a/ot/gromov.py
+++ b/ot/gromov.py
@@ -433,8 +433,7 @@ def fused_gromov_wasserstein(M, C1, C2, p, q, loss_fun='square_loss', alpha=0.5,
 
     where :
     - M is the (ns,nt) metric cost matrix
-    - :math:`f` is the regularization term ( and df is its gradient)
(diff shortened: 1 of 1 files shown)


rflamary:
This seems OK to me. Thank you for the PR.

Will merge in a few days.

PR #139, opened 2020-04-01
[MRG] Fix ordering

AdrienCorenflos:
Fixes https://github.com/rflamary/POT/issues/138

### DIFF
--- a/ot/lp/__init__.py
+++ b/ot/lp/__init__.py
@@ -12,16 +12,16 @@ Solvers for the original linear program OT problem
 
 import multiprocessing
 import sys
+
 import numpy as np
 from scipy.sparse import coo_matrix
 
-from .import cvx
-
+from . import cvx
+from .cvx import barycenter
 # import compiled emd
 from .emd_wrap import emd_c, check_result, emd_1d_sorted
-from ..utils import parmap
-from .cvx import barycenter
 from ..utils import dist
+from ..utils import parmap
 
 __all__ = ['emd', 'emd2', 'barycenter', 'free_support_barycenter', 'cvx',
            'emd_1d', 'emd2_1d', 'wasserstein_1d']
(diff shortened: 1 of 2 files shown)


ot/lp/__init__.py
```diff
@@ -656,7 +656,7 @@ def emd_1d(x_a, x_b, a=None, b=None, metric='sqeuclidean', p=1., dense=True,
     perm_a = np.argsort(x_a_1d)
     perm_b = np.argsort(x_b_1d)
 
-    G_sorted, indices, cost = emd_1d_sorted(a, b,
+    G_sorted, indices, cost = emd_1d_sorted(a[perm_a.flatten()], b[perm_b.flatten()],
```
rtavenar, line 659: I guess you do not have to flatten the permutation indices since they are computed from 1d arrays, or do you ?

test/test_ot.py
```diff
@@ -91,6 +91,44 @@ def test_emd_1d_emd2_1d():
...
+    wass1d = log["cost"]
+    wass1d_emd2 = ot.emd2_1d(u, v, w_u, w_v, metric='sqeuclidean', log=False)
+    wass1d_euc = ot.emd2_1d(u, v, w_u, w_v, metric='euclidean', log=False)
+
+    # check loss is similar
+    np.testing.assert_allclose(wass, wass1d)
+    np.testing.assert_allclose(wass, wass1d_emd2)
+
+    # check loss is similar to scipy's implementation for Euclidean metric
+    wass_sp = wasserstein_distance(u.reshape((-1,)), v.reshape((-1,)))
```
rtavenar, line 123: You forgot the weights here, which is probably why the test fails at the moment

AdrienCorenflos:
Errr, sorry about that. Tbh I ran the asserts in the first comment but not
the test in the PR :)

rflamary:
Hello @AdrienCorenflos,

Thank you for finding the bug and the PR.

In addition to the comments from @rtavenar  be careful to check pep8 on test_ot.py or else the tests will fail.

ot/lp/__init__.py
rtavenar: Could you explain why you flatten here?

Otherwise, LGTM!

PR #155, opened 2020-04-22
debug circle deploy

agramfort:


rflamary:
Hello @agramfort 

Since you are working of this maybe we could and a test in deploy so that it does not fail (or deploy anything) if there is no change in the output of the doc?

rflamary:
yes!

https://pythonot.github.io/auto_examples/index.html

agramfort:
yes success

it won't commit anything if there is no diff. It'a git repo that commits
incremental changes

>

PR #156, opened 2020-04-22
[WIP] Remove notebooks from repo and cleanup doc/readme

rflamary:


agramfort:
all the materials is present in the sphinx gallery examples?

rflamary:
notebooks were identical to the examples and they are available in the doc. I removed the list from the readme because it also requires to be updated.   It lead to an heavy and heavy to pull repo when they were updated.


I think it would be nice to put links directly in the itemize at the top of the redame that point toward examples it will reduce the amount of updating work in my opinion.

docs/source/readme.rst
```diff
@@ -1,64 +1,115 @@
...
-   implementation (requires cupy).
+Source Code (MIT): https://github.com/PythonOT/POT
+
+POT provides the following generic OT solvers (links to examples):
+
+-  `OT Network Simplex
+   solver <https://pythonot.github.io/auto_examples/plot_OT_1D.html>`__
+   for the linear program/ Earth Movers Distance [1] .
+-  `Conditional
+   gradient <https://pythonot.github.io/auto_examples/plot_optim_OTreg.html>`__
```
agramfort, line 21: within sphinx I would not hard code the links but rely on cross refs within sphinx files. Otherwise the old version of the doc will keep linking to new files

rflamary: OK done,

plus if you type make rdoc the readme.md is converted to readme.rst in the doc with relative path (sed magic).

PR #157, opened 2020-04-22
[MRG] build wheels on release

rflamary:
+ build automatically wheels on release 
+ hopefully for manylinux/windows/macosx


.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -U "cython"
+
```
agramfort, line 29: Why is cython not in requirements ?

rflamary: it is!

but the build fails if i dont' do that... This is weird.

README.md
```diff
@@ -86,9 +87,9 @@ You can install the toolbox through PyPI with:
 ```
 pip install POT
 ```
-or get the very latest version by downloading it and then running:
+or get the very latest version by running:
 ```
-python setup.py install --user # for user install (no root)
+pip install git+https://github.com/PythonOT/POT.git # with --user for user install (no root)
```
agramfort, line 92:
```suggestion
pip install -U https://github.com/PythonOT/POT/archive/master.zip # with --user for user install (no root)
```

it will also work on windows without git installed.

PR #158, opened 2020-04-23
[WIP] Issue with sparse emd and adding tests on macos

ncourty:
This PR is intended to solve a bug occuring with the sparse return of coupling matrix in the C version of EMD. This bug was causing a segmentation fault in the OSX distribution, but most importantly was not working with the expected behavior in all cases


rflamary:
I' sad to see sparse emd go but it was nearly useless in practice and since the solver is dens it did not really made sens.

Thank you again @agramfort for caching this.

PR #163, opened 2020-05-01
[WIP] Improved docs and changed scipy version

rtavenar:
I changed the scipy version requirements since version scipy 1.2.1 made my POT crash (cannot remember on which call, sorry, it happened while building the docs) and the issue was fixed when upgrading to scipy 1.3

Apart from that, the main goal of this PR is to homogenize a bit the presentation in the docs.


agramfort:
updating the dep should impact the test suite. Ideally we need a build running minimal dependencies

rtavenar:
Did you mean something like that? And isn't it a problem that we do not test newer versions? It's unclear to me what the best option is.

And also, tests seem not to run for this PR, or maybe Github Actions is down at the moment?

agramfort:
yes something like that.

I would however not run all CIs with minimal deps. I would have only one just to check if backward compat is not broken.

You could then test if the bug your saw on your machine in repoducible on the CI and if not it's not mandatory to bump the version number

my 2c

rtavenar:
OK, Will do

rtavenar:
Good catch @agramfort : it must have been an issue with my local install, since the tests pass on a minimal config with scipy 1.2. I will revert the changes in the `README.md` and keep the test on minimal config (for numpy, scipy, cython and matplotlib), if it's OK for you.

rtavenar:
OK, so [minimal deps lead to failing tests](https://github.com/rtavenar/POT/runs/640482972?check_suite_focus=true).
It seems that with the minimal deps installed, the `autograd` version is old and its `builtins` module does not exist.

I will check that asap (yet, not now).

rflamary:
Hell @rtavenar and thank you for the PR.

numpy 1.16 is ok with me 

I like the idea of having a one test for minimal deps. This will allow us to follow what breaks especially when new features are proposed and then we can increment those minimal deps. 
But then why don't we need to put those in the setup.py then? setup.py is the one used with pip and forcing minimal version would ensure that after install the proper deps are available no?

But now what is killing me is github action and why aren't they reported in the PR! This is a major pain because they run but then i have to go the the fork to have  look.

rflamary:
Also with minimal dependencies maybe we should not install autograd or pymanop.

the tests are designed (or should be) not to fail but to skip in this configuration.

wdyt @agramfort and @rtavenar

rflamary:
yes!

rflamary:
OK @rtavenar,

your new test does not appear yet (until it's on the master nrach i guess) but it failed it seems. This is great because we did not have any visibility of older numpy/scipy before. Here it seems that it fails partly because np.quantile did not exist before numpy 1.15.0 so yes definitely move to 1.16 as minimal dependency.

agramfort:
@rtavenar tell me when to look. I agree with the approach so far.

rtavenar:
Yep @agramfort it's not ready yet, will let you know :)

rflamary:
Looks good to me, @agramfort do you agree?

agramfort:
thx a lot @rtavenar

PR #170, opened 2020-05-05
fix array bounds issue

### LINKED ISSUE #169: unrespected bounds
Hi,

There is a (subtle) bug in EMD_1d: the indices i and j are allowed to reach n and m, therefore these lines have unkown behaviour (the resulting values w_i and w_j are dereferenced from the next memory slot in the machine, therefore it could be any value or even fail depending on compiler).

An easy fix would be to change the while condition to a while True and add break condiditions within the if-else branches.

https://github.com/PythonOT/POT/blob/0a1e8cd6d63b354fe7dfe0439c08417ecf988121/ot/lp/emd_wrap.pyx#L191

https://github.com/PythonOT/POT/blob/0a1e8cd6d63b354fe7dfe0439c08417ecf988121/ot/lp/emd_wrap.pyx#L199

I'll send a pull request if you are happy with this alternative

AdrienCorenflos:
Fixes #169

### DIFF
--- a/ot/lp/emd_wrap.pyx
+++ b/ot/lp/emd_wrap.pyx
@@ -157,12 +157,12 @@ def emd_1d_sorted(np.ndarray[double, ndim=1, mode="c"] u_weights,
         cost associated to the optimal transportation
     """
     cdef double cost = 0.
-    cdef int n = u_weights.shape[0]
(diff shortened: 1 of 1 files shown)


AdrienCorenflos:
this failure is because cur_idx is not incremented after we break

PR #172, opened 2020-05-05
add normalization of distances

HexuanLiu:


rflamary:
Thank you for the pr.

Note that the tests failed because of pep8 violation that can be easily solved with autopep8. I will do a proper code review in the following days but note that the normalization should be an option in the function and be False by default to ensure that it will not break existing code.

ot/dr.py
```diff
@@ -173,7 +183,7 @@ def cost(P):
             for j, xj in enumerate(xc[i:]):
                 xj = np.dot(xj, P)
                 M = dist(xi, xj)
-                G = sinkhorn(wc[i], wc[j + i], M, reg, k)
+                G = sinkhorn(wc[i], wc[j + i], M, reg/regmean[i,j], k)
```
rflamary, line 186: the reg term should be proportional to the distance (since there is a division when computing the K matrix in sinkhorn)

rflamary:
```suggestion
                G = sinkhorn(wc[i], wc[j + i], M, reg*regmean[i,j], k)
```

PR #174, opened 2020-05-13
[MRG] Fix requirements

djsutherland:
As mentioned in https://github.com/conda-forge/pot-feedstock/pull/9#issuecomment-628189995 – the `setup_requires` and `install_requires` seem to be slightly wrong as they are now (and there's no such key as `requires` as far as I can tell).


djsutherland:
It's failing from `flake8` on a file I didn't touch...I'll send a different PR I guess.

djsutherland:
(Rebased onto #176 to avoid the flake8 error.)

rflamary:
I agree this is a much better requirement definition. we were clearly over eager before.

Issue #118, opened 2019-12-25
Cannot install POT on Mac with 'pip install POT'


GitHberChen:
system: MacOS 10.15.1
pip install POT
Looking in indexes: http://mirrors.aliyun.com/pypi/simple/
Collecting POT
  Downloading http://mirrors.aliyun.com/pypi/packages/4e/23/8aba82a55d40ac8c5f4313c1c51804dcb239ee3761b9789db1b621e43c87/POT-0.6.0.tar.gz (794kB)
     |████████████████████████████████| 798kB 393kB/s 
Requirement already satisfied: numpy in /Users/chenlinwei/anaconda3/envs/DL3.7/lib/python3.7/site-packages (from POT) (1.18.0)
Requirement already satisfied: scipy in /Users/chenlinwei/anaconda3/envs/DL3.7/lib/python3.7/site-packages (from POT) (1.1.0)
Requirement already satisfied: cython in /Users/chenlinwei/anaconda3/envs/DL3.7/lib/python3.7/site-packages (from POT) (0.29.2)
Building wheels for collected packages: POT
  Building wheel for POT (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /Users/chenlinwei/anaconda3/envs/DL3.7/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/tmp/pip-install-tfqvon7y/POT/setup.py'"'"'; __file__='"'"'/private/tmp/pip-install-tfqvon7y/POT/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/tmp/pip-wheel-0mnp0bwh --python-tag cp37
       cwd: /private/tmp/pip-install-tfqvon7y/POT/
  Complete output (49 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.macosx-10.7-x86_64-3.7
  creating build/lib.macosx-10.7-x86_64-3.7/ot
  copying ot/unbalanced.py -> build/lib.macosx-10.7-x86_64-3.7/ot
  copying ot/bregman.py -> build/lib.macosx-10.7-x86_64-3.7/ot
  copying ot/plot.py -> build/lib.macosx-10.7-x86_64-3.7/ot
[...]

PR #130, opened 2020-02-28
[MRG] solution to osx issue

ncourty:
This PR is intended to solve an issue preventing POT to be installed on the latest version of OSX using 'pip install pot'

PR #175, opened 2020-05-13
[WIP] only mangle CC/etc if on mac and not told not to

djsutherland:
As mentioned in https://github.com/conda-forge/pot-feedstock/pull/9#issuecomment-628189995 and following up on #130  – always setting `CC=g++` is the wrong thing to do, and also always doing the `CFLAGS` mangling on Mac, regardless of what compilers you're doing/etc, is also the wrong thing to do. This only sets `CC` on Mac, and allows you to bail out of doing any of this (as e.g. conda-forge will want) by setting `POT_LEAVE_CC`.

CC @ncourty


djsutherland:
(Rebased onto #176 to avoid the flake8 error.)

setup.py
```diff
@@ -39,7 +36,9 @@
 
 
 # add platform dependant optional compilation argument
-if sys.platform.startswith('darwin'):
+if sys.platform.startswith('darwin') and not os.environ.get('POT_LEAVE_CC', ''):
+    os.environ["CC"] = "g++"
+    os.environ["CXX"] = "g++"
```
agramfort, line 41: why do we need to do this in the first place? POT does not compile with clang or icc?

djsutherland: I guess it was intended to fix https://github.com/PythonOT/POT/issues/118 (from #130) for people installing with the system compilers on Mac, but I don't really understand why this would be needed.

agramfort:
this is for me not the right fix.

I typically do this when using clang on my mac:

CFLAGS='-stdlib=libc++'  pip install -e .

and not need for a hack in setup.py but maybe there is a simpler way.

>

djsutherland:
I don't know why this would ever be generally needed; my worry is that the current behavior is over-reacting to one user's broken build system. But @ncourty can clarify maybe.

ncourty:
Hello all. Yes, I do not think changing the CC flag is needed anymore. I remember it was an old solution to OSX installation problems before clang becomes the main c++ compiler on OSX. Clearly it is a reminiscence from the past, I guess we can remove it now ?

agramfort:
I propose to remove it and see if CIs break. If not we can remove

rflamary:
good for me

rflamary:
OK now this PR is just removing the flags and it still works in CI. Since @djsutherland basically had to remove it for it to build for conda I guess we should merge it if everyone is OK with it.

ncourty:
ok with this !

djsutherland:
For the record, I don't love the `isysroot` stuff either, and will probably patch it away on conda-forge. But I _think_ it's probably not too bad in the vast majority of situations.

haampie:
Can you maybe make a new release that includes this change? 0.7.0 still has it

haampie:
Or how about a `0.7.1` release that just cherry-picks this PR, that'd be even better.

PR #176, opened 2020-05-13
rename I to img in examples, since flake8 complains now

djsutherland:
The new version of flake8 complains about the variable name `I` now; it didn't before. Rename these to `img`, which is better anyway.


agramfort:
thx @djsutherland

PR #178, opened 2020-05-15
[MRG] Codecov options + remove 2.7

ayushkarnawat:
## What does this PR do?
Instead of specifying the files to ignore within .coveragerc, I believe it is recommended to add those settings inside the `codecov.yml` file instead.

I also removed some unnecessary imports and lines from setup that were no longer used. 

As a side note, for now, I have specified that the codecov bot only comments on pull requests whenever there is a change (current behavior), should we set it to comment even if there are no changes to the coverage (recommend by codecov)?


ayushkarnawat:
Also, since we moved strictly away from python 2.7, should we remove the `externals` module since it is no longer necessary?

setup.py
```diff
@@ -21,73 +19,71 @@
...
+        'Operating System :: POSIX',
+        'Programming Language :: Python',
+        'Programming Language :: C++',
+        'Programming Language :: C',
+        'Programming Language :: Cython',
+        'Topic :: Utilities',
+        'Topic :: Scientific/Engineering :: Artificial Intelligence',
+        'Topic :: Scientific/Engineering :: Mathematics',
+        'Topic :: Scientific/Engineering :: Information Analysis',
+        'Programming Language :: Python :: 2',
```
agramfort, line 82: Python 2 should be removed

agramfort:
Yes external should be removed

ayushkarnawat:
This is probably not the best place to address this, but I think, for PRs, we should always squash the changes rather than merge commit them. This prevents us from logging the whole git history (i.e. minor commits to fix issues addressed by reviewers). By squashing the changes into 1 commit, we only save the history of the changes that matter (i.e. change from the previous pull request). I think it leaves a much cleaner git history. Just my 2c

agramfort:
I disabled "Allow merge commits" in settings

>

ayushkarnawat:
I think there may be some issues with the codecov tool, its not reporting on the coverage correctly. Working on a fix rn

ayushkarnawat:
It seems that the codecov reports failed to upload properly, I think this error is on codecov's end

```
Error: 400 Client Error: Bad Request for url: https://storage.googleapis.com/codecov/v4/raw/2020-05-17/696E8FC9A239519C4D149C953501626B/7e105c15fa17192eac0d3985188fb282652da2d1/46f5ab3b-f76d-4562-b3c9-f31dbe2e6a09.txt?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=GOOG1EQX6OZVJGHKK3633AAFGLBUCOOATRACRQRQF6HMSMLYUP6EAD6XSWAAY%2F20200517%2FUS%2Fs3%2Faws4_request&X-Amz-Date=20200517T143224Z&X-Amz-Expires=10&X-Amz-SignedHeaders=host&X-Amz-Signature=a1384d3f15efc4d169d7344583b78d3037fa3dacaf6f21e48239f07bd5b3b811
```

rflamary:
@ayushkarnawat actually you had the wrong path for gpu it is actually ot/gpu/*.py

Sorry i did not see that earlier. Right now we are at 88% caverage could you fix that please?

ayushkarnawat:
Whoops my bad. Sorry for the late reply. Seems like you already fixed it! Thanks for catching it.

PR #192, opened 2020-06-18
[WIP] OpenMP EMD

### LINKED ISSUE #187: Unable to compute multiple EMD loss
## Describe the bug

Attempting to use the EMD (specifically `ot.emd2`) to compute the loss returns a pickling error. This is likely because the loss computation could not be pickled properly (on MacOS atleast). 

### To Reproduce
1. Run code snippet below

#### Stack trace

```
Traceback (most recent call last):
  File "examples/test_emd2.py", line 25, in <module>
    emd_loss = ot.emd2(source, targets, cost)
  File "/Users/ayushkarnawat/Documents/dev/python_workspace/POT/ot/lp/__init__.py", line 429, in emd2
    res = parmap(f, [b[:, i] for i in range(nb)], processes)
  File "/Users/ayushkarnawat/Documents/dev/python_workspace/POT/ot/utils.py", line 249, in parmap
    p.start()
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/context.py", line 283, in _Popen
    return Popen(process_obj)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/Users/ayushkarnawat/miniconda3/envs/pot/lib/python3.8/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'emd2.<locals>.f'
```

#### Screenshots

N.A.

#### Code sample

```python
import numpy as np

import ot
from ot.datasets import make_1D_gauss as gauss


num_bins = 100
num_targets = 50

# Bin positions
x = np.arange(num_bins, dtype=np.float64)
means = np.linspace(20, 90, num_targets)

# Generate gaussian distributions
source = gauss(num_bins, m=20, s=5)  # m= mean, s= std
targets = np.zeros((num_bins, num_targets))
for i, m in enumerate(means):
    targets[:, i] = gauss(num_bins, m=m, s=5)

# Compute cost matrix and normalize
cost = ot.dist(x.reshape((num_bins, 1)), x.reshape((num_bins, 1)), metric="euclidean")
cost /= cost.max()

# Compute EMD loss
emd_loss = ot.emd2(source, targets, cost)
```

### Expected behavior

The loss between each source and target is computed properly. For the example provided, this should be, 

`[0.0, 0.014426924576813046, 0.0288559930665439, 0.04328574159548381, 0.057715687365961325, 0.07214568550413558, 0.08657569638056341, 0.1010057100978733, 0.1154357243964259, 0.12986573880413424, 0.1442957532306665, 0.15872576766018065, 0.17315578209012886, 0.1875857965201352, 0.20201581095014873, 0.21644582538016313, 0.23087583981017767, 0.245305854240192, 0.2597358686702063, 0.2741658831002206, 0.2885958975302355, 0.30302591196024975, 0.31745592639026415, 0.3318859408202786, 0.346315955250293, 0.36074596968030737, 0.3751759841103219, 0.38960599854033634, 0.40403601297035013, 0.4184660274003606, 0.43289604183033936, 0.44732605626007876, 0.4617560706881306, 0.47618608510525495, 0.4906160994573777, 0.5050461134543907, 0.5194761256704825, 0.5339061296913891, 0.5483360991299938, 0.5627659348379075, 0.5771952970338474, 0.5916231256118063, 0.6060464162679144, 0.6204574579249451, 0.6348383926216666, 0.6491520836171457, 0.6633296042796002, 0.6772576353391107, 0.6907726671326471, 0.7036700128578843]`

- OS (e.g. MacOS, Windows, Linux): MacOS
- Python version: 3.8.2
- How was POT installed (source, `pip`, `conda`): source
- Build command you used (if compiling from source): `python setup.py develop`
- Only for GPU related bugs:
    - CUDA version:
    - GPU models and configuration:
    - Any other relevant information:

```
macOS-10.13.6-x86_64-i386-64bit
Python 3.8.2 (default, May  6 2020, 02:49:43) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.17.4
SciPy 1.4.1
POT 0.7.0
```

ayushkarnawat:
## Types of changes

Python 3.8 no longer supports forking a process (rather it spawns it on a new thread) on MacOS. In ideal conditions -- that is when the code can be trivially pickled to be sent to the individual threads -- this doesn't lead to any problems. However, cython code (which allows the CPP code to be called from within python) cannot be easily pickled. As such, to avoid picking errors and to fix the problems associated with forking a process on windows as well, it is better to use parallel processing in the core CPP code (w/ OpenMP). Resolves #187.

## Checklist

- [ ] The documentation is up-to-date with the changes I made.
- [ ] All tests passed, and additional code has been covered with new tests.

### DIFF
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -92,14 +92,14 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        # install OpenMP (via homebrew)
+        /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+        brew install libomp
         # install (necessary) recipes
         pip install --upgrade pip
         pip install -r requirements.txt
         pip install pytest "pytest-cov<2.6"
         pip install -U "sklearn"
-        # install OpenMP (via homebrew)
-        /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
-        brew install libomp
     - name: Set environment variables
       run: |
         echo '::set-env name=CC::/usr/bin/clang++' # hacky(!!) force c++ compiler
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -96,7 +96,7 @@ jobs:
         /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
         brew install libomp
         # install (necessary) recipes
-        pip install --upgrade pip
+        pip install --upgrade pip setuptools
         pip install -r requirements.txt
         pip install pytest "pytest-cov<2.6"
         pip install -U "sklearn"
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -83,6 +83,7 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: [3.7]
+        env-name: pot
 
     steps:
     - uses: actions/checkout@v1
@@ -93,7 +94,7 @@ jobs:
     - name: Setup miniconda
       uses: goanpeca/setup-miniconda@v1
       with:
-        activate-environment: pot
+        activate-environment: ${{ matrix.env-name }}
         python-version: ${{ matrix.python-version }}
         channels: conda-forge,defaults
         allow-softlinks: true
@@ -102,14 +103,15 @@ jobs:
         use-only-tar-bz2: true
     - shell: bash -l {0}
       run: |
-        # upgrade conda
-        # conda update -n base -c defaults -y conda
+        # update conda
+        conda update -n base -c defaults -y conda
+        # activate env
+        source activate ${{ matrix.env-name }}
         # install (necessary) recipes
-        # python -m pip install --upgrade pip
-        # pip install -r requirements.txt
-        while read requirement; do conda install --yes $requirement; done < requirements.txt
-        conda install -y pytest-cov # hacky, might not be <2.6 
-        # pip install -U "sklearn"
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install pytest "pytest-cov<2.6"
+        pip install -U "sklearn"
         # install OpenMP-enabled C/C++ compilers
         conda install -y "conda-forge::compilers>=1.0.4"
         conda install -y conda-forge::llvm-openmp
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -96,23 +96,23 @@ jobs:
         pip install -r requirements.txt
         pip install pytest "pytest-cov<2.6"
         pip install -U "sklearn"
-        # https://github.com/scikit-learn/scikit-learn/blob/master/doc/developers/advanced_installation.rst
-        # on macOS, install an OpenMP-enabled clang/llvm to set env vars for build
-        # this has to be done using conda since recipes aren't available on PyPI
-        # alternatively, macOS compilers (via homebrew)
+      # https://github.com/scikit-learn/scikit-learn/blob/master/doc/developers/advanced_installation.rst
+      # on macOS, install an OpenMP-enabled clang/llvm to set env vars for build
+      # this has to be done using conda since recipes aren't available on PyPI
+      # alternatively, macOS compilers (via homebrew)
         xcode-select --install
         /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
         brew install libomp
     - name: Install POT
       run: | # build in verbose mode (to check for the presence of the -fopenmp flag in the compiler commands) 
         pip install --verbose --no-build-isolation -e .
-      env: |
-        CC: /usr/bin/clang
-        CXX: /usr/bin/clang++
-        CPPFLAGS: "${{ CPPFLAGS }} -Xpreprocessor -fopenmp"
-        CFLAGS: "${{ CFLAGS }} -I/usr/local/opt/libomp/include"
-        CXXFLAGS: "${{ CXXFLAGS }} -I/usr/local/opt/libomp/include"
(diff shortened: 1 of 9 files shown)


ayushkarnawat:
Sorry for the delay in getting this started. There are things that still need to be completed before a review. In particular,

- [ ] Proper compile flags for OpenMP across all OSes (setup.py)
- [ ] Remove parmap (not needed anymore)
- [ ] Don't require OpenMP dependency?? (Code can still compile/run, but EMD computation will be a lot slower)
  - [ ] Modify bonneel code
  - [ ] Tests to check if OpenMP is supported (w/in pkg)
  - [ ] Fail fast OpenMP code (checks for support)
  - [ ] Modify CI pipeline to run tests that require/don't require OpenMP (add env variables to ensure proper compilation of CPP code)
- [ ] Update cython min version to 0.28.5 (for py35)
- [ ] Rename files, i.e. `EMD_wrapper.cpp` -> `emd.cpp` (or something similar)

.github/workflows/build_tests.yml
```diff
@@ -71,37 +71,56 @@ jobs:
...
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        # install OpenMP (via homebrew)
+        /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+        brew install libomp
```
agramfort, line 97: can you also keep a macos build without openmp so we can check it works on a fresh machine too? thx

rflamary: yes indeed i would keep the classical build on macosx and add a build/test with openmp

ayushkarnawat:
@agramfort @rflamary Some insight/pointers would be appreciated.

Although the package is being installed properly on all platforms (with correct flags it seems), it is for some reason not recognizing `get_omp_num_threads` within the [Linux](https://github.com/PythonOT/POT/runs/793330571#step:7:21) and [Mac](https://github.com/PythonOT/POT/runs/793330560#step:8:27) builds, respectively. These work flawlessly on my local build, so can't quite pin down the issue with the CI builds. Interestingly, it works on [Windows](https://github.com/PythonOT/POT/pull/192/checks?check_run_id=793330552#step:6:6) without a problem. 

**NOTE**: The tests need to be modified to address the new changes, hence why some tests fail.

.github/workflows/build_tests.yml
ayushkarnawat: Yes, I was planning on doing this (just after I fixed the issues with the full build)

rflamary:
Looks to me as if it builds bu do not link openmp.

There is an option for the cython builder called `extra_link_args` that can be used for that I guess. I would try it with the same value that you add to compile args.

ayushkarnawat:
Thanks for the quick reply. I'll give that a try!

agramfort:
did you look how scikit-learn does it ? https://github.com/scikit-learn/scikit-learn/blob/master/setup.py#L137

it took a lot of time to @jeremiedbb to make it work on scikit-learn.

ayushkarnawat:
@agramfort I did look into it, but would it be premature optimization? I say this because the implementation in scipy looks like it is meant to handle many submodules that contain C/C++ code to compile properly. Although I agree that approach is better (i.e. cleaner and more robust for future additions of C/C++ submodule(s)), we currently only have one module where C/C++ code is being used, hence why I thought it might have been easier to define how to cythonize the EMD module in `setup.py` directly. 

Alternatively, since we have to check if OpenMP is supported within a user's system (so that we can run EMD computation with/without OpenMP support), we may have to end up mimicking how scipy is approaching this since the (a) compile flags would change accordingly. and (b) the cythonization would change depending on whether the system could support OpenMP.

agramfort:
@ayushkarnawat are you suggesting that making sure it works on systems with or without openmp support automatically is premature optimization?

ayushkarnawat:
Perhaps I was not clear enough. What I meant to say was the way scipy handles checking if OpenMP is supported in the the `scipy/_build_utils` module. From looking at the code, it seems this was done to:
1. check if a user's system supports OpenMP (through a small test program)
2. to define cythonization of submodules within each respective submodule's `setup.py`.

I agree that (1) should also be done in POT since checking if the user's system supports OpenMP is important (as mentioned previously). This allows us to add the proper openmp linker to the compiler so that the generated `.so` file can use parallel acceleration if the user's system supports it, and fallback into single-threaded use, if that is not possible.  

What I meant by "premature optimization" was concerning (2). This is because it seems like it was meant to support cythonization of multiple C++ src files contained within different submodules, which we (currently) do not have. To be pedantic, I don't think we need to define the cythonization in the fashion that scipy does it. I guess we simply can forgo that portion, if you agree.  

Let me know what you think and if this makes sense.

agramfort:
as long as we check if a user's system supports OpenMP (through a small
test program)
and allow POT to work on a system without openmp it's fine for me.

>

rflamary:
Thank you @ayushkarnawat for this work, You have motivated us to look into openMP also. 

we have another PR that has done the OpenMP implementation (actually both single thread and openMP) and currently build on all OS in #260 so I will close this PR.

ayushkarnawat:
Very cool, apologies for not finishing this earlier; got busy with other things and sort of forgot about it. I'm happy to actually seeing this actually being done. :)

Issue #198, opened 2020-07-12
line_search_armijo throws exception


panpan2:
## Describe the bug
scipy.optimize.linesearch.scalar_search_armijo can return `alpha=None`. In optim.py:line_search_armijo as part of a previous issue [https://github.com/PythonOT/POT/issues/184](https://github.com/PythonOT/POT/issues/184), we restrict the value of `alpha` to be at most 1 with `min(1, alpha)` but this throws an Exception when `alpha` is `None`.

### Expected behavior
Return `None` instead of throwing an Exception

Output of the following code snippet:
```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import ot; print("POT", ot.__version__)

'''
Darwin-19.2.0-x86_64-i386-64bit
Python 3.7.7 (default, May  6 2020, 04:59:01)
[Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.18.1
SciPy 1.2.1
POT 0.7.0
'''
```

### Additional context

https://github.com/PythonOT/POT/blob/acfff525e043c8733936e8e99367543edd28822b/ot/optim.py#L72

rflamary:
Closed in #199

Issue #184, opened 2020-05-20
line_search_armijo returns alpha > 1


octavian-ganea:
line_search_armijo can return alpha > 1 because that's what the scipy function was doing. However, this affects the cg() function used in both GW and FGW functions and returns transport plans with negative entries when armijo=True.

rflamary:
This is indeed a problem. 

If you want to do a PR to handle this case (basically return min(1,alpha)) you are welcome. Else we will get around to it in the following days.

rflamary:
Closed in #189

PR #199, opened 2020-07-12
Check if alpha is not None when restricting it to be at most 1

panpan2:
## Types of changes
scipy.optimize.linesearch.scalar_search_armijo can return alpha=None ([function definition](https://github.com/scipy/scipy/blob/master/scipy/optimize/linesearch.py#L689)). In optim.py:line_search_armijo as part of a previous issue #184, we restrict the value of alpha to be at most 1 with min(1, alpha) but this throws an Exception when alpha is None. In this PR, we just check the value of `alpha` is not `None` when making the comparison with `1`.

Issue: [https://github.com/PythonOT/POT/issues/198](https://github.com/PythonOT/POT/issues/198)

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] All tests passed, and additional code has been covered with new tests.

### DIFF
--- a/ot/optim.py
+++ b/ot/optim.py
@@ -69,7 +69,7 @@ def line_search_armijo(f, xk, pk, gfk, old_fval,
     alpha, phi1 = scalar_search_armijo(
         phi, phi0, derphi0, c1=c1, alpha0=alpha0)
 
-    return min(1, alpha), fc[0], phi1
+    return min(1, alpha) if alpha is not None else 1, fc[0], phi1
 
 
 def solve_linesearch(cost, G, deltaG, Mi, f_val,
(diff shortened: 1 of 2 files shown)


ot/optim.py
```diff
@@ -69,7 +69,7 @@ def phi(alpha1):
     alpha, phi1 = scalar_search_armijo(
         phi, phi0, derphi0, c1=c1, alpha0=alpha0)
 
-    return min(1, alpha), fc[0], phi1
+    return min(1, alpha) if alpha is not None else 1, fc[0], phi1
```
rflamary, line 72: could you put the test in another line with a small comment about what is necessary?

It is a bit hard to follow in the return

rflamary: Also I guess you have an example where the original function failed. Could you do a simple no regression test it test_optim.py to check that this problem never re-appears again? It feels as if this function always needs tuning ;)

rflamary: Hello @panpan2 could you please take into account my comment so that we can merge your PR?

panpan2: Updated! :)

panpan2: @rflamary any other changes needed to merge this?

Issue #126, opened 2020-02-24
Infeasible solution for EMD but inputs are in the simplex?


alsuhr-c:
**Describe the bug**
A call to the EMD solver is raising a `UserWarning` about the problem being infeasible (`Check that a and b are in the simplex`). I dug into the code and this seems to be raised only when a value in one of the two inputs is less than zero. However, I am printing the sum, max, and min of the inputs and don't see any odd behavior (sum to one, max values not greater than 1, min values not less than zero; inputs are the same size; type is `np.float64`). 

Could this be an issue with underflow?

**To Reproduce**
Steps to reproduce the behavior: using python3.7:

```python
import ot
import torch
import random
import numpy as np

# Get distributions
random_dist = torch.softmax(torch.tensor([random.random() * 1000 for _ in range(625)]), dim=0).numpy()
random_dist2 = torch.softmax(torch.tensor([random.random() * 1000 for _ in range(625)]), dim=0).numpy()
print('--- First dist ---')
print('Shape: %s' % random_dist.shape)
print('Sum: %s' % np.sum(random_dist))
print('Max: %s' % random_dist.max())
print('Min: %s' % random_dist.min())
print('--- Second dist ---')
print('Shape: %s' % random_dist2.shape)
print('Sum: %s' % np.sum(random_dist2))
print('Max: %s' % random_dist2.max())
print('Min: %s' % random_dist2.min())

# Create distance matrix
coords = np.reshape(np.asarray(np.meshgrid(np.linspace(0, 24, 25), np.linspace(0, 24, 25)))[[1, 0], :, :].transpose((1, 2, 0)), [-1, 2])
tiled = np.tile(coords[:, np.newaxis, :], [1, coords.shape[0], 1])
ttiled = tiled.transpose((1, 0, 2))
distances = np.array(np.linalg.norm(tiled - ttiled, axis=2), order='C')

# Compute EMD
ot.emd(random_dist, random_dist2, distances)
```

Example output:

```
--- First dist ---
Shape: 625
Sum: 1.0
Max: 0.42820615
Min: 0.0
--- Second dist ---
Shape: 625
Sum: 1.0
Max: 0.96281725
Min: 0.0
/home/.../venv/lib/python3.7/site-packages/ot/lp/__init__.py:113: UserWarning: Problem infeasible. Check that a and b are in the simplex
  result_code_string = check_result(result_code)
```

**Expected behavior**
I would expect that this error is not thrown as all of the values in both distributions are nonnegative.

[...]

PR #217, opened 2020-11-12
change precision EPSILON in C code

### LINKED ISSUE #93: Not in simplex -- two sets of largely different sizes
I am trying to calculate the EMD of two sets. When one set has a few hundred entries and the other has only 2, the EMD calculation fails and returns Problem Infeasible.

** SEE BELOW COMMENT FOR FIXED SCRIPT ** 

**Expected behavior**
Should return EMD around 1, instead says that the sets spherEng1 and pencilEnergy are not in the simplex

**Screenshots**
Here is comparing the EMDs calculated for less densely tiled to most densely tiled (number of particles = number of segments) with the two element set
image

**Desktop (please complete the following information):**
 - OS: [MacOSX]
 - Python version [3.6]
- POT installed with pip

>>> import platform; print(platform.platform())
Darwin-16.7.0-x86_64-i386-64bit
>>> import sys; print("Python", sys.version)
('Python', '2.7.15 |Anaconda, Inc.| (default, Dec 14 2018, 13:10:39) \n[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]')
>>> import numpy; print("NumPy", numpy.__version__)
('NumPy', '1.15.4')
>>> import scipy; print("SciPy", scipy.__version__)
('SciPy', '1.1.0')
>>> import ot; print("POT", ot.__version__)
('POT', '0.5.1')

ncourty:
## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

Problem INFEASIBLE cast where it should not happen.

Fixes https://github.com/PythonOT/POT/issues/126

Fixes #93 

The minimal code to repdroduce the bug is now working.
A user tested on its own data and now the problem is also working

## Checklist

- [X] The documentation is up-to-date with the changes I made.
- [X] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [X] All tests passed, and additional code has been covered with new tests.

### DIFF
--- a/ot/lp/network_simplex_simple.h
+++ b/ot/lp/network_simplex_simple.h
@@ -34,8 +34,8 @@
 #endif
 
 
-//#define EPSILON 2.2204460492503131e-14
-#define EPSILON 1.1e-7
+#define EPSILON 2.2204460492503131e-15
+//#define EPSILON 1.1e-7
 #define _EPSILON 1e-8
 #define MAX_DEBUG_ITER 100000
 
@@ -1508,7 +1508,7 @@ namespace lemon {
 			if( retVal == OPTIMAL){
                 for (int e = _search_arc_num; e != _all_arc_num; ++e) {
                     if (_flow[e] != 0){
-                        if (abs(_flow[e]) > EPSILON)
+                        if (fabs(_flow[e]) > _EPSILON)
                             return INFEASIBLE;
                         else
                             _flow[e]=0;
--- a/ot/lp/network_simplex_simple.h
+++ b/ot/lp/network_simplex_simple.h
@@ -34,7 +34,8 @@
 #endif
 
 
-#define EPSILON 2.2204460492503131e-15
+//#define EPSILON 2.2204460492503131e-14
+#define EPSILON 1.1e-7
 #define _EPSILON 1e-8
 #define MAX_DEBUG_ITER 100000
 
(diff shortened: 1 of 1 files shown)


ot/lp/network_simplex_simple.h
```diff
@@ -34,7 +34,8 @@
 #endif
 
 
-#define EPSILON 2.2204460492503131e-15
+//#define EPSILON 2.2204460492503131e-14
+#define EPSILON 1.1e-7
```
agramfort, line 38: maybe you use FLT_EPSILON from cmath but it's probably good enough

ot/lp/network_simplex_simple.h
```diff
@@ -35,6 +35,7 @@
 
 
 #define EPSILON 2.2204460492503131e-15
+//#define EPSILON 1.1e-7
```
rflamary, line 38: You can remove now

ot/lp/network_simplex_simple.h
```diff
@@ -1507,7 +1508,7 @@ namespace lemon {
 			if( retVal == OPTIMAL){
                 for (int e = _search_arc_num; e != _all_arc_num; ++e) {
                     if (_flow[e] != 0){
-                        if (abs(_flow[e]) > EPSILON)
+                        if (fabs(_flow[e]) > _EPSILON)
```
rflamary, line 1511: maybe add a comment to say that it was switched to _EPSILON due to numerical problems and refer to the Issue Number

Issue #221, opened 2020-12-21
Numpy AxisError in ot.bregman.sinkhorn_epsilon_scaling


rubencart:
## Describe the bug

I am getting an error when using `ot.bregman.sinkhorn_epsilon_scaling`, I get the same error for any argument values I've tried. Method `ot.bregman.sinkhorn_stabilized`, with the same arguments, works.

### To Reproduce
Minimal example:
```python
import numpy as np
import ot
import ot.gpu
num_samples, dim = 1000, 256
xs = np.random.randn(num_samples, dim) - 1
xt = np.random.randn(num_samples, dim) + 1
a, b = np.ones((num_samples,)) / num_samples, np.ones((num_samples,)) / num_samples
M = ot.dist(xs, xt, metric='euclidean')
M /= M.max()
reg = 1
W = ot.sinkhorn2(a, b, M, reg, log=True, numItermax=1000, method="sinkhorn_epsilon_scaling")
```
`ot.__version__` gives 0.7.0, and I'm running python 3.7.6 with cython version 0.29.21.

Error message:
```python
---------------------------------------------------------------------------
AxisError                                 Traceback (most recent call last)
<ipython-input-17-e02d798141ea> in <module>
      9 M /= M.max()
     10 reg = 1
---> 11 W = ot.sinkhorn2(a, b, M, reg, log=True, numItermax=1000, method="sinkhorn_epsilon_scaling")

/export/home1/NoCsBack/hci/rubenc/miniconda3/envs/genv/lib/python3.7/site-packages/ot/bregman.py in sinkhorn2(a, b, M, reg, method, numItermax, stopThr, verbose, log, **kwargs)
    232         return sinkhorn_epsilon_scaling(a, b, M, reg, numItermax=numItermax,
    233                                         stopThr=stopThr, verbose=verbose,
--> 234                                         log=log, **kwargs)
    235     else:
    236         raise ValueError("Unknown method '%s'." % method)

/export/home1/NoCsBack/hci/rubenc/miniconda3/envs/genv/lib/python3.7/site-packages/ot/bregman.py in sinkhorn_epsilon_scaling(a, b, M, reg, numItermax, epsilon0, numInnerItermax, tau, stopThr, warmstart, verbose, print_period, log, **kwargs)
    937             transp = G
    938             err = np.linalg.norm(
--> 939                 (np.sum(transp, axis=0) - b)) ** 2 + np.linalg.norm((np.sum(transp, axis=1) - a)) ** 2
[...]

PR #224, opened 2021-01-03
solve issue #221

ncourty:
## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

Problem with function 'sinkhorn_epsilon_scaling'

solve issue #221 

The patch was tested on the code used for reproducing the bug as given in issue #221 

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x ] All tests passed, and additional code has been covered with new tests.


ot/bregman.py
```diff
@@ -955,9 +966,23 @@ def get_reg(n):  # exponential decreasing
...
-        return G, log
+        if output_value:
+            output = sinkhorn_stabilized(a, b, M, regi,
+                                         numItermax=numInnerItermax, stopThr=1e-9,
+                                         warmstart=(alpha, beta), verbose=False,
+                                         print_period=20, tau=tau, log=False)
+            return output, log
+        else:
+            return G, log
     else:
```
rflamary, line 977: is that an else for the loop? is that necessary? it is not covered by the tests  also.

rflamary (commented):
Just a small question

ot/bregman.py
rflamary: Still  +0.84%  coverage is awesome (also the bug was in this percent ;) )

ot/bregman.py
```diff
@@ -919,8 +927,11 @@ def get_reg(n):  # exponential decreasing
     while loop:
 
         regi = get_reg(cpt)
-
-        G, logi = sinkhorn_stabilized(a, b, M, regi,
+        # b.flatten() is here for the case when 'sinkhorn_epsilon_scaling' has been called
+        # from sinkhorn2. In this casen b is a 2D array of size [..,1] and 'sinkhorn_stabilized'
+        # does not return a valid coupling, but rather transport cost.
+        # This is a patch to issue #221
+        G, logi = sinkhorn_stabilized(a, b.flatten(), M, regi,
```
AdrienCorenflos, line 964: I was looking at the same issue earlier today, I don't think this is an ok way of doing this tbh. Flattening may create further bugs. I'm of the opinion of delegating the role of making sure the inputs are one dimensional to sinkhorn2 instead of implicitly converting them (several times) in the loop.

AdrienCorenflos: Otherwise, I also had in mind that the transport plan sinkhorn functions could return batched transport plans if n_hists > 1, but this may require more work...

AdrienCorenflos (commented):
I'm not sure that the fix is a good one, it assumes a certain form of structure without enforcing it

rflamary:
Closing this PR since the option epsilon_scaling was removed from sinkhorn2

PR #238, opened 2021-04-15
[MRG] build wheels with arm64 (raspberry PI)

rflamary:
+ Build automatically weekly all wheels even for arm (via emulation with QEMU)
+ Add optional 'no ci' and 'no pep8' flags in commit message to disable temporarilly ci and pep8 check


rflamary:
Yeah so now we can build wheels for aarch64 (that is for Raspberry PI and android devices). 

But since it uses emulation it basically takes forever to build them (45 minutes instead of 10).

The wheels are built only on master and when forced in PR with 'build wheels' in the commit message so it should not be seen during PR but it will inrease the chance that the wheels build will fail on master branch (I kind of hat those red X).

WDYT @ncourty @agramfort ?

agramfort:
sklearn builds / uses ARM on a weekly basis as it's too costly / slow.

I would not make ARM part of the main CI run for every commit. It's not
sustainable.

>

rflamary:
OK now we should have  weekly build for all the wheels (that can be triggered if needed) and pus on master for the fast ones.

PR #254, opened 2021-06-01
[WIP] Add Wasserstein GAN and debug memory leak

rflamary:
This PR introduce the following changes

- [x] New example for Wasserstein GAN
- [x] Correct memory leak in ot.emd2 for torch backend
- [x] Speedup color Domain Adaptation examples for a quicker compilation of the doc

 The memory leak was due to the creation of a new class for each call to ot.emd2 that used torch tensors in their scope which meant that tose tensors could not be de-allocated by the garbage collector. It has been solved by using a torch function that takes both inputs output value and gradients as input bu provides gardients only for the inputs (see set_gradients(self, val, inputs, grads) function on torch backend)

### DIFF
--- a/examples/backends/plot_wass2_gan_torch.py
+++ b/examples/backends/plot_wass2_gan_torch.py
@@ -55,7 +55,8 @@ import matplotlib.pyplot as pl
 import ot
 import torch
 
-# %% Data generation
+# %%
+# Data generation
 # ---------------
 
 torch.manual_seed(1)
@@ -72,9 +73,10 @@ def get_data(n):
     return x
 
 
-##############################################################################
+# %%
 # Plot data
 # ---------
+
 # %% plot the distributions
 x = get_data(500)
 pl.figure(1)
@@ -83,12 +85,11 @@ pl.title('Data distribution')
 pl.legend()
 
 
-##############################################################################
+# %%
 # Generator Model
 # ---------------
-# %% define the MLP model
 
-
+# define the MLP model
 class Generator(torch.nn.Module):
     def __init__(self):
         super(Generator, self).__init__()
@@ -105,10 +106,9 @@ class Generator(torch.nn.Module):
         output = self.fc3(output)
         return output
 
-##############################################################################
+# %%
 # Training the model
 # ------------------
-# %%
 
 
 G = Generator()
@@ -160,8 +160,9 @@ pl.title('Wasserstein distance')
 pl.xlabel("Iterations")
 
 
-# %% Plot trajectories of generated samples along iterations
-# ----------------------------------------------------------
+# %%
+# Plot trajectories of generated samples along iterations
+# -------------------------------------------------------
 
 
 pl.figure(3, (10, 10))
@@ -178,10 +179,10 @@ for i in range(9):
     if i == 0:
         pl.legend()
 
-##############################################################################
+# %%
 # Generate and visualize data
 # ---------------------------
-# %%
+
 size_batch = 500
 xd = get_data(size_batch)
 xn = torch.randn(size_batch, 2)
--- a/examples/backends/plot_wass2_gan_torch.py
+++ b/examples/backends/plot_wass2_gan_torch.py
@@ -51,16 +51,15 @@ and Statistics (Vol. 108).
 
 from torch import nn
 import numpy as np
-import matplotlib.pylab as pl
+import matplotlib.pyplot as plt
 import ot
 import torch
 
 
 ##############################################################################
 # Data generation
-# -------------
+# ---------------
 
-#%% Data generation
 torch.manual_seed(1)
 ncirc = 6
 sig = 0.1
--- a/examples/backends/plot_wass2_gan_torch.py
+++ b/examples/backends/plot_wass2_gan_torch.py
@@ -77,7 +77,7 @@ def get_data(n):
 # Plot data
 # ---------
 
-# %% plot the distributions
+# plot the distributions
 x = get_data(500)
 pl.figure(1)
 pl.scatter(x[:, 0], x[:, 1], label='Data samples from $\mu_d$', alpha=0.5)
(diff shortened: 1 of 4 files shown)


examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+# Author: Remi Flamary <remi.flamary@polytechnique.edu>
+#
+# License: MIT License
+
+# sphinx_gallery_thumbnail_number = 2
+
+
+from torch import nn
+import numpy as np
+import matplotlib.pylab as pl
```
agramfort, line 54:
```suggestion
import matplotlib.pyplot as plt
```
this is way more standard

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+import ot
+import torch
+
+
+##############################################################################
+# Data generation
+# -------------
```
agramfort, line 61:
```suggestion
# ---------------
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+
+#%% Data generation
```
agramfort, line 63:
```suggestion
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+torch.manual_seed(1)
+ncirc = 6
+sig = 0.1
+n = 100
+d = 2
+p = 2
```
agramfort, line 69: can we avoid one letter variables?

n_samples
n_dims
n_features

?

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+##############################################################################
+# Data generation
+# -------------
+
+#%% Data generation
+torch.manual_seed(1)
+ncirc = 6
+sig = 0.1
```
agramfort, line 66: can you can this sigma or std? sig was not obvious to me?

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+        output = self.relu(output)
+        output = self.fc3(output)
+        return output
+
+##############################################################################
+# Training the model
+# ------------------
+
+
+#%%
```
agramfort, line 120: if you want to use such vscode / spyder cells you can use them instead of the #### line above.
Sphinx gallery understand them.

Yet I think it should be 
```suggestion
# %%
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,206 @@
...
+G = Generator()
+optimizer = torch.optim.RMSprop(G.parameters(), lr=0.001)
+
+# number of iteration and size of the batches
+niter = 500
```
agramfort, line 125:
```suggestion
n_iter = 500
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,194 @@
...
+# License: MIT License
+
+# sphinx_gallery_thumbnail_number = 3
+
+
+from torch import nn
+import numpy as np
+import matplotlib.pyplot as pl
+import ot
+import torch
```
agramfort, line 56:
```suggestion
import numpy as np
import matplotlib.pyplot as pl
import torch
from torch import nn
import ot
```

import from more standard to local.

also I note that you don't like my plt :)

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,194 @@
...
+
+# %% Data generation
+# ---------------
```
agramfort, line 59:
```suggestion
# %%
# Data generation
# ----------------
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,194 @@
...
+
+torch.manual_seed(1)
+sigma = 0.1
+n_dims = 2
+n_features = 2
+
+
+def get_data(n):
```
agramfort, line 67: n -> n_samples?

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,194 @@
...
+    angle = c * 2 * np.pi
+    x = torch.cat((torch.cos(angle), torch.sin(angle)), 1)
+    x += torch.randn(n, 2) * sigma
+    return x
+
+
+##############################################################################
+# Plot data
+# ---------
+# %% plot the distributions
```
agramfort, line 78:
```suggestion
# %%
# Plot data
# ---------
#
```

examples/backends/plot_wass2_gan_torch.py
```diff
@@ -0,0 +1,194 @@
...
+pl.figure(1)
+pl.scatter(x[:, 0], x[:, 1], label='Data samples from $\mu_d$', alpha=0.5)
+pl.title('Data distribution')
+pl.legend()
+
+
+##############################################################################
+# Generator Model
+# ---------------
+# %% define the MLP model
```
agramfort, line 89: idem

examples/backends/plot_wass2_gan_torch.py
rflamary: github is bugging  I alreday commited some changes but nothing is appearing here. will take into account the other comments and push again

And no I don't like plt, i'm opinionated on this one ;)

rflamary:
https://635-71472695-gh.circle-artifacts.com/0/dev/auto_examples/backends/plot_wass2_gan_torch.html#sphx-glr-auto-examples-backends-plot-wass2-gan-torch-py

Simple 2D Wasserstein GAN in 1 min and 30 seconds.

PR #258, opened 2021-06-08
[MRG] Update example GAN to avoid the 10 minute CircleCI limit

rflamary:
All in the title ;)

Also added an animation for the trajectory of the generated samples


rflamary:
https://672-71472695-gh.circle-artifacts.com/0/dev/auto_examples/backends/plot_wass2_gan_torch.html#sphx-glr-auto-examples-backends-plot-wass2-gan-torch-py

Now with animation (also wuicker to compile)

agramfort (approved):

Issue #277, opened 2021-09-08
Implement set_gradient for JAX backend


AdrienCorenflos:
One can implement it easily by using compensating sums.

PR #278, opened 2021-09-08
Add set_gradients method for JAX backend.

AdrienCorenflos:
## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ X ] New feature (non-breaking change which adds functionality)

The `set_gradient` is possible in JAX.

https://github.com/PythonOT/POT/issues/277

Added  a modified unittest for JAX.

## Checklist

- [ X ] The documentation is up-to-date with the changes I made.
- [ X ] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [ X ] All tests passed, and additional code has been covered with new tests.

### DIFF
--- a/ot/backend.py
+++ b/ot/backend.py
@@ -287,16 +287,16 @@ class JaxBackend(Backend):
             return jnp.array(a).astype(type_as.dtype)
 
     def set_gradients(self, val, inputs, grads):
-        # no gradients for jax because it is functional
+        from jax.flatten_util import ravel_pytree
+        val, = jax.lax.stop_gradient((val,))
 
-        # does not work
-        # from jax import custom_jvp
-        # @custom_jvp
-        # def f(*inputs):
-        #     return val
-        # f.defjvps(*grads)
-        # return f(*inputs)
+        ravelled_inputs, _ = ravel_pytree(inputs)
+        ravelled_grads, _ = ravel_pytree(grads)
 
+        aux = jnp.sum(ravelled_inputs * ravelled_grads) / 2
+        aux = aux - jax.lax.stop_gradient(aux)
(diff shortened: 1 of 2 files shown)


rflamary:
This is awesome!  that's why you need people who know the framework for multiple backend.

Could you please also check the you can compute gradients for emd2? by updating this test?
https://github.com/PythonOT/POT/blob/master/test/test_ot.py#L84

AdrienCorenflos:
Well, no you won't be able to. You are casting your tensors to numpy, and JAX tracing is not compatible with this... In order to use JAX you would need to dispatch the call to the host by using a callback instead (see https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html), and only after this set the gradients. In practice this means that the backends should essentially have a kind of pycall method.

AdrienCorenflos:
> practice this means that the backends should essentially have a kind of pycall method.

Note that this is feasible, but it's a bit more work than simply setting gradients, and requires some preliminary design thinking.

rflamary:
OK i'm sorry i vnever understoiod the subtleties of jax. But if what you implemented did not define the gradient for a given variable, what is its use?

AdrienCorenflos:
It does replace the gradient within JAX code, but it does not allow to bypass the tracing. The problem of emd2 is not with the gradient, it's with the fact that you have an operation that JAX can't trace in the middle.

rflamary:
Note that We use the set_gradient in emd2 exactly to bypass the need for numpy arrays and it works for torch.  This is why i tried to define a new function with @custom_jvp (and call it) in set gradients but i wasn't a good strategy.

PR #336, opened 2022-01-17
[MRG] Implement Sinkhorn in log-domain for WDA

jakubzadrozny:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

The current implementation of the WDA dimensionality reduction runs into numerical issues (nans and infs) with small values of the regularization parameter (reg). This PR re-implements the sinkhorn algorithm in log-domain, which allows to use WDA with much smaller reg parameter values.

A new regression test was added that runs WDA with small reg, the current master version fails it and this PR passes.

## PR checklist

### DIFF
--- a/ot/dr.py
+++ b/ot/dr.py
@@ -30,16 +30,25 @@ def dist(x1, x2):
     return x1p2.reshape((-1, 1)) + x2p2.reshape((1, -1)) - 2 * np.dot(x1, x2.T)
 
 
-def sinkhorn(w1, w2, M, reg, k):
-    r"""Sinkhorn algorithm with fixed number of iteration (autograd)
+def logsumexp(M, axis):
+    r"""Log-sum-exp reduction compatible with autograd (no numpy implementation)
     """
-    K = np.exp(-M / reg)
-    ui = np.ones((M.shape[0],))
-    vi = np.ones((M.shape[1],))
+    amax = np.amax(M, axis=axis, keepdims=True)
+    return np.log(np.sum(np.exp(M - amax), axis=axis)) + np.squeeze(amax, axis=axis)
+
+
+def sinkhorn(w1, w2, M, reg, k):
+    r"""Sinkhorn algorithm in log-domain with fixed number of iteration (autograd)
+    """
+    Mr = -M / reg
+    ui = np.zeros((M.shape[0],))
+    vi = np.zeros((M.shape[1],))
+    log_w1 = np.log(w1)
+    log_w2 = np.log(w2)
     for i in range(k):
-        vi = w2 / (np.dot(K.T, ui))
-        ui = w1 / (np.dot(K, vi))
-    G = ui.reshape((M.shape[0], 1)) * K * vi.reshape((1, M.shape[1]))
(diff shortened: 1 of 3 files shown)


ot/dr.py
```diff
@@ -30,16 +30,25 @@ def dist(x1, x2):
     return x1p2.reshape((-1, 1)) + x2p2.reshape((1, -1)) - 2 * np.dot(x1, x2.T)
 
 
+def logsumexp(M, axis):
+    r"""Log-sum-exp reduction compatible with autograd (no numpy implementation)
+    """
+    amax = np.amax(M, axis=axis, keepdims=True)
+    return np.log(np.sum(np.exp(M - amax), axis=axis)) + np.squeeze(amax, axis=axis)
```
rflamary, line 37: Why are you using `keepdims=True` and then `squeeze`? It should it work without both?

rflamary:
Hello @jakubzadrozny and thank you for the PR.

I just have a question in my code review but as soon as you change the code or discuss why (and the tests pass) we can merge.

rflamary:
Also pleas remember to add your contribution to the REALEASES.md file as a New feature and your name/email as co-contributor on top of the dr.py file

ot/dr.py
jakubzadrozny: `M - amax` (inside of `exp`) won't work without `keepdims=True` when reducing on `axis=1`. As an example:
```
>>> M = np.random.randn(10, 20)
>>> amax = np.amax(M, axis=1)
>>> M - amax
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (10,20) (10,) 
```
while
```
>>> amax = np.amax(M, axis=1, keepdims=True)
>>> M - amax
```
works fine.

rflamary: ok you are right

jakubzadrozny:
Hi @rflamary. Thanks for your comments, I've answered your question and pushed another commit.

rflamary:
Hello again @jakubzadrozny and thank you for your changes.

I nearly merged and then noticed something: log sinkhorn is much slower.

Your implementation
https://1255-71472695-gh.circle-artifacts.com/0/dev/auto_examples/others/plot_WDA.html#compute-wasserstein-discriminant-analysis
takes 30 sec on the example

while the release version 
https://pythonot.github.io/auto_examples/others/plot_WDA.html#compute-wasserstein-discriminant-analysis
takes 9sec to run

I think it would be nice to add your ling implementation to the classical sinkhorn and add a parameter to WDA function to sleect classical or log space (with a message in the doc to say which to choose).

Could you please do that?

jakubzadrozny:
Sorry, I honestly did not expect it to be that much slower. I've added the `sinkhorn_method` parameter to `wda` with the previous implementation as default. I tried to follow `ot.bregman.sinkhorn` with the naming convention and also the doc message, let me know if you'd like that changed.

rflamary:
Well losumexp is exactly 3 times more complex than np.dot ;). That is the reason I keep vanilla sinkhorn as default even if for deep leraning and 32 bits you always need to go toward log stabilization. 

The PR looks good I'm merging it thank you for your contribution.

PR #341, opened 2022-01-28
[MRG] Add weak OT solver

rflamary:
## Types of changes

* Add weak OT solver
* Add tests for solver and backends
* Add example for weak OT in doc

## PR checklist

### DIFF
--- a/ot/weak.py
+++ b/ot/weak.py
@@ -1,0 +1,123 @@
+"""
+Weak optimal ransport solvers
+"""
+
+# Author: Remi Flamary <remi.flamary@polytehnique.edu>
+#
+# License: MIT License
+
+from .backend import get_backend
+from .optim import cg
+import numpy as np
+
+__all__ = ['weak_optimal_transport']
+
+
+def weak_optimal_transport(Xa, Xb, a=None, b=None, verbose=False, log=False, G0=None, **kwargs):
+    r"""Solves the weak optimal transport problem betwen two empirical distributions
+
+
+    .. math::
+        \gamma = \mathop{\arg \min}_\gamma \quad \|X_a-diag(1/a)\gammaX_b\|_F^2
+
+        s.t. \ \gamma \mathbf{1} = \mathbf{a}
+
+             \gamma^T \mathbf{1} = \mathbf{b}
+
+             \gamma \geq 0
+
+    where :
+
+    - :math:`X_a` :math:`X_b`  are the sample matrices.
+    - :math:`\mathbf{a}` and :math:`\mathbf{b}` are the sample weights
+
+
+    .. note:: This function is backend-compatible and will work on arrays
+        from all compatible backends.
+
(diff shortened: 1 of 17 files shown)


ot/weak.py
```diff
@@ -0,0 +1,123 @@
...
+    ot.optim.cg : General regularized OT
+    """
+
+    nx = get_backend(Xa, Xb)
+
+    Xa2 = nx.to_numpy(Xa)
+    Xb2 = nx.to_numpy(Xb)
+
+    if a is None:
+        a2 = nx.ones((Xa.shape[0]), type_as=Xa) / Xa.shape[0]
```
ncourty, line 94: type_as should be related to Xa2 or directly in numpy ?

test/test_weak.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_weak_ot_bakends(nx):
+    # test weak ot solver for different backends
+    n = 50
+    rng = np.random.RandomState(0)
+
+    xs = rng.randn(n, 2)
+    xt = rng.randn(n, 2)
+    u = ot.utils.unif(n)
```
ncourty, line 44: makes me think there is no type_as for unif ?

ncourty (changes requested):
thx for the nice work rémi !

ot/weak.py
```diff
@@ -0,0 +1,123 @@
...
+# License: MIT License
+
+from .backend import get_backend
+from .optim import cg
+import numpy as np
+
+__all__ = ['weak_optimal_transport']
+
+
+def weak_optimal_transport(Xa, Xb, a=None, b=None, verbose=False, log=False, G0=None, **kwargs):
```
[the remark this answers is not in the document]
rflamary: the API of this function is closer to empirical sinkhorn
https://pythonot.github.io/gen_modules/ot.bregman.html?highlight=empirical_sinkhorn#ot.bregman.empirical_sinkhorn

where the weights are usually uniform so this order makes more sens

ot/weak.py
rflamary: good catch

test/test_weak.py
rflamary: added even though it is the only function in teh public API with this parameter (type cannot be inferred from integer)

ot/weak.py
```diff
@@ -0,0 +1,123 @@
...
+
+             \gamma \geq 0
+
+    where :
+
+    - :math:`X_a` :math:`X_b`  are the sample matrices.
+    - :math:`\mathbf{a}` and :math:`\mathbf{b}` are the sample weights
+
+
+    .. note:: This function is backend-compatible and will work on arrays
```
ncourty, line 35: yes  but maybe you can add a note about  the fact that all  the computations happen on the CPU in numpy mode

ncourty (approved):

PR #347, opened 2022-02-11
[WIP] Build POT against oldest-supported-numpy

davidghiurco:
## Types of changes

Build / setup:
* Compiling against the oldest possible `numpy` version possible given the platform, while also still allowing `numpy>=1.20` to be used
* Removing legacy code workarounds to build-time dependency problems which were fixed by the inclusion of a `pyproject.toml` in https://github.com/PythonOT/POT/pull/293
* Set `python_requires` to `>=3.7`. Python 3.6 support was dropped in [POT release `0.8.1`](https://github.com/PythonOT/POT/releases/tag/0.8.1). This configuration is also necessary to help `oldest-supported-numpy` properly determine what the oldest version of `numpy` it should use during setup.

## Motivation and context / Related issue

Address #346 

## PR checklist

- [X] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [ ] The documentation is up-to-date with the changes I made (check build artifacts).
- [ ] All tests passed, and additional code has been **covered with new tests**.
- [ ] I have added the PR and Issue fix to the [**RELEASES.md**](RELEASES.md) file.

PR #349, opened 2022-02-22
[MRG] Build POT against oldest-supported-numpy (local PR)

### LINKED ISSUE #346: Change pyproject.toml to require the oldest possible numpy version to prevent future backwards-breaking numpy ABI changes
## Describe the bug

Currently the [0.8.1.0 release pyproject.toml](https://github.com/PythonOT/POT/blob/be0ed1dcd0b65ec804e454f1a2d3c0d227c0ddbe/pyproject.toml) has:

```
[build-system]
requires = ["setuptools", "wheel", "numpy>=1.20", "cython>=0.23"]
build-backend = "setuptools.build_meta"
```
The problem with `numpy>=1.20` is that for source installations, as well as wheel builds, this will actually use the *latest* (at build time) version of `numpy`. This can have the unintended consequence of making the resulting `POT` build compatible *only* with that latest numpy version (and newer).

This is because builds against numpy are *forward compatible*, but not *backward compatible*. (Source: [numpy docs](https://numpy.org/devdocs/user/depending_on_numpy.html#:~:text=NumPy's%20ABI%20is%20forward%20but,but%20not%20with%20older%20versions.)). What this means for `POT` is that it should be compiled against the *oldest* `numpy` version that it supports. In our current case, this would be `numpy==1.20.0`, not whatever is the newest version at the time of compilation with `numpy>=1.20` in `pyproject.toml`.

Addressing this problem will prevent future bug reports when POT is inevitably built against `numpy>=1.22`, which has a backwards-breaking ABI change to 1.20.x & 1.21.x (see https://numpy.org/doc/stable/release/1.22.0-notes.html#compatibility-notes, and https://github.com/numpy/numpy/pull/19479)

There is a meta-package developed by numpy devs called [oldest-supported-numpy](https://pypi.org/project/oldest-supported-numpy/) which is meant to be used *exclusively* in `pyproject.toml` for the purposes of *building/compiling* against numpy. Newer versions of numpy can still be used as *runtime* dependencies without breaking binary compatibility. 

### To Reproduce
1.build current codebase from latest release `0.8.1.0` from source, it will be built against `numpy>=1.22.x` (which was released December 31st, 2021).
2. Attempt to use the compiled package on a platform running `numpy==1.20.x` or `numpy==1.21.x`, you will get something similar to: `ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject`

### Expected behavior
POT should be built against the oldest `numpy` package it supports. This does not prevent using newer numpy versions as *runtime* dependencies. 

- OS (e.g. MacOS, Windows, Linux): MacOS Monterey, Version 12.1
- Python version: 3.8+ (any python version that supports `numpy 1.22.x` is affected). 
- How was POT installed (source, `pip`, `conda`): pip & source, but this problem affects *future* wheel POT builds. 
- Build command you used (if compiling from source):

rflamary:
## Types of changes

this PR is a local (branch) version of #347 that will actually run the wheel building.

Fixes #346

## PR checklist

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -28,7 +28,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install -U "cython"
 
     - name: Install cibuildwheel
       run: |
@@ -37,7 +36,6 @@ jobs:
     - name: Build wheels
       env:
         CIBW_SKIP: "pp*-win* pp*-macosx* cp2* pp* cp36*" # remove pypy on mac and win (wrong version)
-        CIBW_BEFORE_BUILD: "pip install numpy cython"
       run: |
         python -m cibuildwheel --output-dir wheelhouse
 
@@ -81,7 +79,6 @@ jobs:
     - name: Build wheels
       env:
         CIBW_SKIP: "pp*-win* pp*-macosx* cp2* pp* cp*musl* cp36*" # remove pypy on mac and win (wrong version)
-        CIBW_BEFORE_BUILD: "pip install numpy cython"
         CIBW_ARCHS_LINUX: auto aarch64 # force aarch64 with QEMU
         CIBW_ARCHS_MACOS: x86_64 universal2 arm64
       run: |
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -27,7 +27,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
 
     - name: Install cibuildwheel
       run: |
@@ -63,8 +62,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
-        pip install -U "cython"
 
     - name: Install cibuildwheel
       run: |
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ setup(
     data_files=[],
     setup_requires=["oldest-supported-numpy", "cython>=0.23"],
     install_requires=["numpy>=1.16", "scipy>=1.0"],
-    python_requires=">=3.7",
+    python_requires=">=3.6",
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -91,7 +92,6 @@
         'Topic :: Scientific/Engineering :: Mathematics',
         'Topic :: Scientific/Engineering :: Information Analysis',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
```
davidghiurco, line 95:
```suggestion
        'Programming Language :: Python :: 3.6',
        'Programming Language :: Python :: 3.7',
```

.github/workflows/build_wheels.yml
```diff
@@ -81,7 +76,6 @@ jobs:
     - name: Build wheels
       env:
         CIBW_SKIP: "pp*-win* pp*-macosx* cp2* pp* cp*musl* cp36*" # remove pypy on mac and win (wrong version)
```
davidghiurco, line 78:
```suggestion
        CIBW_SKIP: "pp*-win* pp*-macosx* cp2* pp* cp*musl*" # remove pypy on mac and win (wrong version)
```

davidghiurco (changes requested):
Adding suggestions to reconcile build restrictions with `python_requires=">=3.6"` in setup.py

davidghiurco:
@rflamary I noticed you might want to apply the changes made in `build_wheels.yml` to `build_wheels_weekly.yml`: https://github.com/PythonOT/POT/blob/master/.github/workflows/build_wheels_weekly.yml#L42-L43

davidghiurco:
And if you do apply those changes, you might want to add to release notes that Python 3.6 support is back.

rflamary:
We are not adding python3.6 support back because python 3.6 is not supported and we should not promote old unsupported code. Actually the wheels are not builded for 3.6 because it is not supported anymore and the CI does not test 3.6. We are just allowing the user to install it even if it is not supported.

I'll update the weekly the wheels in an other PR.

PR #377, opened 2022-05-12
[WIP] Silence std::cerr calls inside of python

stanleyjs:
## Types of changes
Adds flags to disable std::cerr reporting in the cpp backend. Adds parameter `ot.backend.use_cerr` as parameter and `ot.backend.get_cerr()` getter to enable/disable `cerr` reporting inside of python. The parameter defaults to False.

We are using POT inside of jupyter notebook and running into a problem that hits its iter limits. The warnings flood our console. Because the backend code is calling std::cerr directly, rather than passing the error string up to the python stderr stream, traditional context management based redirection (or simply killing sys.stderr inside of python)  does not work.

Have tested the basic use case and enable/disabling the cerr. Works in ipython on 3.9.7.

## PR checklist

- [x ] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [x ] The documentation is up-to-date with the changes I made (check build artifacts).
- [ x] I have added the PR and Issue fix to the [**RELEASES.md**](RELEASES.md) file.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,6 +10,8 @@
 
 - Fixed an issue where we could not ask TorchBackend to place a random tensor on GPU
   (Issue #371, PR #373)
+- Fixed an issue where hitting iteration limits would be reported to stderr by std::cerr regardless of Python's stderr stream status.
+
 
 ## 0.8.2
 
--- a/ot/lp/network_simplex_simple.h
+++ b/ot/lp/network_simplex_simple.h
@@ -1432,9 +1432,7 @@ namespace lemon {
             // Execute the Network Simplex algorithm
             while (pivot.findEnteringArc()) {
                 if(max_iter > 0 && ++iter_number>=max_iter&&max_iter>0){
-                    char errMess[1000];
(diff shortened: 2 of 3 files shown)


rflamary:
Hello @stanleyjs and thank you for your PR. I agree that using `std::cerr`  in the cpp code was a bad move and I've been meaning to change that for a while so your PR is very interesting.

Still I would have done it in a different way. I think we should remove all those printing in the cpp instead of adding a new parameter and use proper python warning when using the `result_code` in the cython function  `emd_c`. This mean that you will be able to desactivate the warnings but they will still be printed by default (and this time properly). It would be nice to implement this and a test that checks that the warning is printed for a very small number of max iteration. Are you OK with my proposition?

stanleyjs:
@rflamary you mean to just strip out all the std::cerr calls?
I agree that that is cleaner, I just didn't want to remove them because it wasn't clear to me if you wanted the c++ to be useful outside of the python context. In that case, perhaps the c++ error reporting would be useful. 

I'll strip it all out and rebase.

rflamary:
yep strip out all you like : the cpp is only for wrapping with cython here

bonus here is we don't add an undocumented parameter to the toolbox

stanleyjs:
@rflamary the newest push has this integrated.

PR #454, opened 2023-04-08
[MRG] Efficient Discrete Multi Marginal Optimal Transport

x12hengyu:
## Types of changes

This introduce DEMD modules, a file `demd.py` contains all the vanilla modules of Efficient Discrete Multi Marginal Optimal Transport Regularization. 
Also includes two examples, `examples/others/plot_demd_1d.py` and `examples/others/plot_demd_gradient_minimize.py`

Add new methods for [Efficient Discrete Multi Marginal Optimal Transport Regularization](https://openreview.net/forum?id=R98ZfMt-jE), paper on ICLR 2023.

Example `plot_demd_1d.py` uses two 1d Gaussian Distribution as test data and compares computing time with LP.
Example `plot_demd_gradient_minimize.py` compares loss between demd minimizes using gradient decent and lp method.
Also tested three functions separately in `test_demd.py` for sanity checks.

## PR checklist


rflamary:
Thanks for the PR. We will do a code reveiw s soon as possible.

x12hengyu:
> Thanks for the PR. We will do a code reveiw s soon as possible.

Thank you for considering!

x12hengyu:
> Thanks for the PR. We will do a code reveiw s soon as possible.

Dear POT Team,

I hope you're well. Just a quick reminder about our group's pull request (#454) from two weeks ago. I understand the team's been busy, but we'd appreciate your feedback when you have a moment. If you need clarification or have questions, feel free to reach out. We are eager to make any necessary adjustments!

Best regards,
Xizheng Yu

ot/demd.py
```diff
@@ -0,0 +1,233 @@
...
+        the objective function evaluated at x. It may also return a matrix of
+        shape (d, n) representing the gradient of the objective function
+        with respect to x, and/or any other dual variables needed for the
+        optimization algorithm. The signature of this function should be:
+        `f(x, d, n, return_dual_vars=False) -> float`
+        or
+        `f(x, d, n, return_dual_vars=True) -> (float, ndarray, ...)`
+    x : ndarray, shape (d, n)
+        The initial point for the optimization algorithm.
+    d : int
```
rflamary, line 178: those shapes can be infered from x they should not be passed as parameters

rflamary (changes requested):
Hello and thanks for the PR,

I did a quick code review. The PR needs major changes so that I can better understand what the new solvers are doing (you cannot ask people to read the paper to use a function, it should be described in the documentation). It is a bit unclear to me if you provide an EMD, multimarginal solver, a barycenter estimator (especially since you function takes only one array as input) so please clarify this in the code and maybe the PR description.

What needs to be clarified and changed:
+ Move all solvers to a submodule better named more clear such as `ot.lp.discrete_emd` .
+ Please respect POT API: use M for ground loss, a,A for distributions (on the simplex) and all parameters names. All function names can be long but must describe precisely what the function does. For instance if a function computes a barycenter then it should be similarly named to other solvers and be called the same way.
+ Add to the documentation of all function, the optimization problems solved in math environment using same names as other POT functions.
+ Take into account all the small comments below.

I know this is a lot of work but we need to ensure that the code fits well in POT and is easy to use and understand by users/conributors.

x12hengyu:
> Hello and thanks for the PR,
> 
> I did a quick code review. The PR needs major changes so that I can better understand what the new solvers are doing (you cannot ask people to read the paper to use a function, it should be described in the documentation). It is a bit unclear to me if you provide an EMD, multimarginal solver, a barycenter estimator (especially since you function takes only one array as input) so please clarify this in the code and maybe the PR description.
> 
> What needs to be clarified and changed:
> 
> * Move all solvers to a submodule better named more clear such as `ot.lp.discrete_emd` .
> * Please respect POT API: use M for ground loss, a,A for distributions (on the simplex) and all parameters names. All function names can be long but must describe precisely what the function does. For instance if a function computes a barycenter then it should be similarly named to other solvers and be called the same way.
> * Add to the documentation of all function, the optimization problems solved in math environment using same names as other POT functions.
> * Take into account all the small comments below.
> 
> I know this is a lot of work but we need to ensure that the code fits well in POT and is easy to use and understand by users/conributors.

Dear POT team,

Thank you for thoughtful and detailed code review. We apologize for the conflicts and problems right now. We will resolve them one by one and make sure it fits well in POT as well as contributors. We appreciate your time and effort. Thank you.

Best

x12hengyu:
Dear POT team,

This pull request update contains significant changes and improvements to our Efficient Discrete Multi Marginal Optimal Transport Regularization implementation. We've worked on enhancing the overall usability and readability following previous suggestions. We appreciate your previous feedback, which guided these revisions. We look forward to your feedback on these updates.

Best Regards,
Xizheng Yu

In summary, here are the key points of the update:

Major changes:
- Merged `greedy_primal_dual` and `demd` to `ot.lp.discrete_mmot`, named `demd_minimize` to `ot.lp.discrete_mmot_converge`.
- Changed parameter names and return values to follow POT's API standard.
- Merge two examples together, removing time and loss compare (discuss using text discription), and add barycenter plots.
- Add more detailed mathematic proves and explanations in documentation.

Comments resolved below:
- The new solvers should be in `ot.lp.discrete_emd` or something else more descriptive
    - Fixed. Move our method under `ot.lp` and named as `discrete_mmot` stands for multi marginal optimal transport
- the example should not need to transpose the data. it means that the API for the implemented function is not good (it should retrun the smae thing as ot.lp.barycenter)
    - For `ot.lp.barycenter`, input distributions A is defined `A (np.ndarray (d,n))`, but stacked data is shaped (n, d), that's why we added transpose here. We have moved transpose to initial generation of distributions A to avoid confusions. This is the same approach as other examples in POT such as [1D Wasserstein barycenter: exact LP vs entropic regularization](https://pythonot.github.io/auto_examples/barycenters/plot_barycenter_lp_vs_entropic.html#sphx-glr-auto-examples-barycenters-plot-barycenter-lp-vs-entropic-py). We also modified both `discrete_mmot_converge` and `discrete_mmot` methods so their input parameter `A` takes same shape (d, n) input as POT style.
- where is the barycener? Objective is nice but the barycenetr shoudl be rtruned
    - `discrete_mmot` denotes a single iteration in the converging calculation. Since we are not defining a learning rate here, we cannot return updated all distributions. But we have updated distributions in each iteration with `A_new = nx.reshape(A, (d, n)) - grad * lr` in converging method, and returned final barycenters/distributions at the end.
- why only plot thr time? pleade also plot the barycenter.
    - Time removed. Added.
- Plot the data
    - Fixed.
- what is all this? sphinx-gallery can print the output of the function properly (assuming it is using print)
    - Removed. We were trying to compare the time efficency between two methods. Therefore I used stdout to retreive the loss and time data from the output.
- not clear what is the comparaison, the baryenter estimation?
    - Our method converges all the distributions rather than calculating a single center for all distributions, the comparasion is intending to show two different methods works on the same distributions.
- the functiuon should reurn data already well formated
    - Fixed. The function is removed following previous comments.
- No need for a function if you just call it oncen alaos visualize the data
    - Removed.
- this s a weird name for a function, if it computes a barycenter then just call it `discrete_emd_barycenter`
    - Fixed. Named as `discrete_mmot_converge`
- also do `bary, ~ = ot.demd_minimize(ot.demd, data, d, n, vecsize, niters=3000, lr=0.00001)` to avoid betting the barycenter later
    - Fixed, added log as an additional return value following the POT API.
- you must stor eand return the loss instead of parsing the text output of the function...
    - Fixed. But we only can get one final loss from log instead all losses through iterations. We will add comment comparasions for the time difference.
- same comment as above
    - Fixed. Added transpose when create A.
- `weights=np.ones(d)/d` or `weights=ot.unif(d)`
    - Fixed.
- if you solve the same problem, the loss should be the same or similar. Why is it so different.
    - They are related problems, but the methods we used are different, which can result in different losses. MMOT problems focuses on identifying a joint distribution such that the marginals are defined by the input distributions over which we wish to measure the dissimilarity. The resulting loss is different from dmmot because barycenter only considers pairwise distances between points, while dmmot takes all possible combinations of points into accounts. 
- you need to plot both baryceneters. Are they the same? if not why? Maybe a discussion here explaining the method vs standard LP
    - We added plots for both barycenters. They are not the same. It is because of the difference between two optimization problems. The main difference is MMOT can handle an arbitrary number of input distributions efficiently and accurately, while standard barycenter methods are limited to two or three input distributions.
- add more detzails heren alaos the references papers
    - added.
- $ is not compatible with sphinx, please use proper math notations. Also what is aa? what is its relation with the p discussed above.
    - Removed $. And we merged `greedy_primal_dual` to `discrete_mmot` for better structure and naming. aa refered to A in that context.
- are those the input distributions? how many of them?
    - Yes, these are input distributions. Changed to A to follow POT API.
- this function needs to foloow POT API. look at https://pythonot.github.io/all.html#ot.emd or https://pythonot.github.io/all.html#ot.emd2.
    - Fixed. Changed `x` to `A`, `aa` to `A`.
- hare you should express the optimization problem that the functio solves with proper maths and noations (agai see ot.emd), define what is generalized EMD, If it is a multimarginal then please explain it
    - The generalized EMD is a higher-dimensional extension to normal EMD that measures the distance between distributions. We defined a method and added mathematic explanation for this in `dist_monge`. 
- the function should return an OT plan (following ot.emd API) and a log dictionary with other information.
    - Followed this schema in `discrete_mmot_converge`
- the output of the function should be of the same type as the input of the function
- should use nx.sum
    - Fixed, used `nx.sum` and `nx.dot`.
- desctibe more precisely what you are solving. is it a barycenter?
    - it is a related problem to barycenter. In MMOT, the goal is to find a joint probability distribution that matches the marginal distributions of multiple input distributions. 
- bad naming, at least discrete_emd or discrete_emd2 if teh function return emd without the plan . if you give the function an empirical distribution then you shoud also put it in the name. Finally emd is computed beteen twoi distibutions so why is there only on naumpy arrya here?
    - Naming changed. We are solving a different problem than emd. EMD normally between two distributions, but generalized EMD extends to higher-dimension.
- those shapes can be inferred from x they should not be passed as parameters
    - Removed unnecessary parameters that can be inferred.

x12hengyu:
Dear POT team,

I hope this message finds you well. I have recently completed work on the bug fix for workflow checks, and I believe it is now ready for review. I understand everyone has busy schedules, but I would appreciate if you could review this pull request and approve a workflow check at your convenience. Please feel free to raise any questions or concerns. Thank you in advance for your time and patient.

Best regards,
Xizheng Yu

rflamary:
We will do that, thanks for your work it is appreciated.

ot/lp/dmmot.py
```diff
@@ -0,0 +1,319 @@
...
+        except Exception:
+            pass
+
+    dualobj = nx.sum([nx.dot(A[:, i], arr) for i, arr in enumerate(dual)])
+
+    log_dict = {'A': xx,
+                'primal objective': obj,
+                'dual': dual,
+                'dual objective': dualobj}
+
```
rflamary, line 182: here it would be nice to retrun a loss withe the gtradeints defined properly so that it can be used in pytorch  with standard gradient decsnet algorithms. To do that you can use the backend function set_gradients that define the forward/backward relations .
An example of its use can be fnd here:
https://github.com/PythonOT/POT/blob/9076f02903ba2fb9ea9fe704764a755cad8dcd63/ot/lp/__init__.py#L529

ot/lp/dmmot.py
```diff
@@ -0,0 +1,319 @@
...
+    if log:
+        return obj, log_dict
+    else:
+        return obj
+
+
+def discrete_mmot_converge(
```
rflamary, line 189: same here I would use `monge_mmot_1dgrid_optimize`to stet clearly what the function does. i also need more discussion about why one would optimize all distributions  together, you use is as some kind of "barycenter" since they all converge to a given distribution but be clear that it is not a barycenetr in the traditional OT sens.

test/test_dmmot.py
```diff
@@ -0,0 +1,51 @@
...
+    np.random.seed(1234)
+    n = 4
+    a1 = ot.datasets.make_1D_gauss(n, m=20, s=5)
+    a2 = ot.datasets.make_1D_gauss(n, m=60, s=8)
+    A = np.vstack([a1, a2])
+    x = np.arange(n, dtype=np.float64).reshape((n, 1))
+    return A.T, x
+
+
+def test_discrete_mmot():
```
rflamary, line 22: since you implemented the function with backends, you should run the tests on arrays from other backends, you can do that by adding a parameter nx to the test function that will be automatically run with all available backends.

examples/others/plot_d-mmot.py
```diff
@@ -0,0 +1,150 @@
...
+print('Obj\t: ', dmmot_obj)
+
+
+# %%
+# Compare Barycenters in both methods
+# ---------
+pl.figure(1, figsize=(6.4, 3))
+for i in range(len(barys)):
+    if i == 0:
+        pl.plot(x, barys[i], 'g-*', label='Discrete MMOT')
```
rflamary, line 74: it would be nice to see visually if you converged by plotting all the individual distributions (seems like you did because your "barycenter" ). maybe you could call it "Monge MMOT minimization" instead of discrete MMOT?

ot/lp/dmmot.py
```diff
@@ -0,0 +1,319 @@
...
+        https://www.sciencedirect.com/ science/article/pii/0166218X93E0121E.
+        Workshop on Discrete Algoritms.
+    """
+
+    return max(i) - min(i)
+
+
+def discrete_mmot(A, verbose=False, log=False):
+    r"""
+    Compute the discrete multi-marginal optimal transport of distributions A.
```
rflamary, line 62: you should explain clearly ere that you suppose that the support of the distributions are supposed integers on the real line, this will be suggested by the new function name but it needs to be stated clearly.

test/test_dmmot.py
rflamary, line 22: It would be nice to have  a test comparing the loss returned by your solver with two marginals et the exact OT solver with absolute ground metric since they should be equivalent no?.

rflamary (changes requested):
Hello,

Thanks for all the work and the major modifications in your code, it is now in my opinion in a much better shape and we are approaching a merge. 

Still I have a few important modifications now that I better understand what your proposed solvers actually do.  I know it can get tiresome but POT is becoming very large and we need to be very careful with positioning new API and solvers.

My understanding is that they are MMOT solvers for marginal distribution that have a support on a regular 1D grid and using a specific ground metric (max-min that has monge property) that corresponds to the absolute value loss/EMD ground metric with two marginals.

This is very interesting and I want it in POT but it also means that we cannot use the very general names of the method that you chose in your published paper because it would lead to much confusion for POT users. So I proposed more precise names below that describe I think better what is solved and done. I provide more specific comments below (for instance having mathematical object follow the name of function parameters).

On a more scientific question I am surprised by the shape or your "barycenter" that is actually from my understanding a convergence point for the minimization of your MMOT formulation. It seems to be very similar to L2 (average) barycenters especially compared to the LP solution. Do you have an intuition why?

x12hengyu:
Dear POT team,

I hope you are doing well! This pull request update contains changes and improvements based on the previous review. We appreciate your previous feedback, which guided these revisions. We look forward to your feedback on these updates.

Best Regards,
Xizheng Yu

In summary, here are the key points of the update:

- Changed function names and answer comments based on suggestions.
- Regarding the similarity of the shape of our "barycenter" to the L2 (average) barycenters: this is likely due to the Monge cost, for simple 1-D examples. I do not believe that the paper describes theory on the relationship of this Monge cost based barycenter to those obtained using the L2 cost. Our result has verified with cvx's gradient and proved to have the same result (cvx compare is available in [paper's repo](https://github.com/ronakrm/demd/blob/main/synth_and_fair_exps/demd/emd_cvxopt.py))


Comments resolved below:
Q) you should compare it to the l2 (np.mean) barycenter because your barycenter looks very similar
    A) changed compare barycenter from lp to l2 (not sure if we should keep both, but l2 does similar to ours result)

Q) both the objective value ad the norm of the gradient increase at the end which is very surprising since it is supposed to be a gradient descent no?
- same here
    A) We added lr decay in `dmmot_monge_1dgrid_optimize` method to control the step size when we approach the minimized obj.
- list of integer indexes...
    - Fixed.

Q) dist_monge is a very generic for a very specific ground cost that indeed has monge property. This one MMOT ground cost with monge property not the only one. dist_monge_max_min for instanec is better
    A) Fixed name.

Q) again this function name is too general, frm the name it looks like a general MMOT solver when in practice it applies only on regular 1D grids (for the marginals) and with a very specific ground metric. I suggest `monge_mmot_1dgrid_loss` that describes the loss for mmot with monge ground cost on a regular 1D grid (and why we only give the distributions weights to the function). It is a mouthfull but we need precise descriptions when creating new functison in a general purpose OT toolbox
     A) Thanks for the detailed explanation on naming suggestion. Since the method is a discrete MMOT problem with Monge costs, we would like to suggest to use a similiar name `dmmot_monge_1dgrid_loss` tentatively. Since we have “d” distributions, so the grid is “d” dimensional, and our “n” is the discretization/number of bins.

Q) state the ground cost instead of using "generalized Monge" that requires to read your paper. I understand the OT plan is indeednet of which Monge cost but you return the loss for a fixed one.
    A) Fixed.

Q) use alpha instead of p to maje the link with the input A of the function
    A) Fixed in all mathematic discriptions.

Q) Do not use x for the OT plan, we use either or a bold matrix T for plan in POT. x is already used for support position (that are integers i in your ase) and we need unified API/notations
    A) Fixed with `\gamma`

Q) here it would be nice to retrun a loss withe the gtradeints defined properly so that it can be used in pytorch with standard gradient decsnet algorithms. To do that you can use the backend function set_gradients that define the forward/backward relations . An example of its use can be fnd here:
    A) Added inside `dmmot_monge_1dgrid_loss`, followed by https://github.com/PythonOT/POT/blob/release0.9/ot/backend.py#L1689

Q) same here I would use monge_mmot_1dgrid_optimizeto stet clearly what the function does. i also need more discussion about why one would optimize all distributions together, you use is as some kind of "barycenter" since they all converge to a given distribution but be clear that it is not a barycenetr in the traditional OT sens.
    A) We would like to suggest a similar naming `dmmot_monge_1dgrid_optimize` for this method. **Discussion:** The main advantage here is the computation cost grows exactly linearly, and can even be faster if some distributions are “interior” to the others. This means that in the worst case we are computing roughly the same amount of “things” as the barycenter approaches, but in average and best cases we “skip” distributions that aren’t important to the computation of the d-dimensional cost. At each iteration step, only those on the “boundary” are being moved. Here's the figure regarding to move the boudary for your reference: ref-mmot

Q) since you implemented the function with backends, you should run the tests on arrays from other backends, you can do that by adding a parameter nx to the test function that will be automatically run with all available backends.
    A) Added nx test. For our algorithm, since it involves multiple tensor modifications, tensorflow's immutable feature conflicts with our usage. Meanwhile, PyTorch's mutiple methods requires conversion from list to tensor. We decide to use simple conversion at the start and end of each method between nx and np, and use np for the algorithm calculation.

Q) it would be nice to see visually if you converged by plotting all the individual distributions (seems like you did because your "barycenter" ). maybe you could call it "Monge MMOT minimization" instead of discrete MMOT?
    A) A plot for comparing all individual distributions has added, but they are really close due to the method (like we stated, every distribution can be view as a "barycenter"). The tentative naming is `dmmot_monge_1dgrid_optimize`.

Q) you should explain clearly ere that you suppose that the support of the distributions are supposed integers on the real line, this will be suggested by the new function name but it needs to be stated clearly.
    A) Added to the documentation.

Q) It would be nice to have a test comparing the loss returned by your solver with two marginals et the exact OT solver with absolute ground metric since they should be equivalent no?.
    A) Added to test

rflamary (changes requested):
Thank you @x12hengyu for all those changes. I know it was a lot of work but it looks good and more maintainable in the future.

I think the contribution is nearly there but there remains a small problem with the backend line where the gradient is set (see below).

Once this is done we can merge , Good work

I will wait for this and do a new release so this should be shortly available in the stable version.

x12hengyu:
Dear POT team,

I hope you are doing well! I have fixed the gradient problem. Thanks for your time and patient in reviewing our work in past months. We really appreciate!

Best Regards,
Xizheng Yu

ot/lp/dmmot.py
```diff
@@ -0,0 +1,344 @@
...
+        Nauk SSSR, 37:227-229, 1942.
+
+    See Also
+    --------
+    ot.lp.dmmot_monge_1dgrid_optimize : Optimize the d-Dimensional Earth
+    Mover's Distance (d-MMOT)
+    """
+
+    nx = get_backend(A)
+    A_copy = nx.copy(A)
```
rflamary, line 140:
```suggestion
    A_copy = A
```

We nee to keep teh original input to have a gradient

rflamary (approved):

PR #459, opened 2023-04-14
[MRG] add the sparsity-constrained optimal transport funtionality and example

liutianlin0121:
## Types of changes
Add a new optimal transport functionality and an example.

"Sparsity-constrained optimal transport" is a variant of the optimal transport problem that provides direct control over the number of non-zero values allowed in the optimal plan. This formulation may be of interest to users of POT. For more information, please refer to the following paper: https://openreview.net/forum?id=yHY9NbQJ5BP.

The tests check that the marginal constraints of the optimal plans are approximately satisfied, and that the sparsity constraints are satisfied. 

## PR checklist

### DIFF
--- a/ot/sparse.py
+++ b/ot/sparse.py
@@ -1,0 +1,229 @@
+"""
+Sparsity-constrained optimal transport solvers.
+
+Implementation of :
+Sparsity-Constrained Optimal Transport.
+Tianlin Liu, Joan Puigcerver, Mathieu Blondel.
+In Proc. of AISTATS 2018.
+https://arxiv.org/abs/1710.06276
+
+[50] Liu, T., Puigcerver, J., & Blondel, M. (2023).
+Sparsity-constrained optimal transport.
+Proceedings of the Eleventh International Conference on
+Learning Representations (ICLR).
+"""
+
+# Author: Tianlin Liu <t.liu@unibas.ch>
+#
+# License: MIT License
+
+
+import numpy as np
+import ot
+from .backend import get_backend
+
+
+class SparsityConstrained(ot.smooth.Regularization):
+    """ Squared L2 regularization with sparsity constraints """
+
+    def __init__(self, max_nz, gamma=1.0):
+        self.max_nz = max_nz
+        self.gamma = gamma
+
+    def delta_Omega(self, X):
+        # For each column of X, find entries that are not among the top max_nz.
+        non_top_indices = np.argpartition(
+            -X, self.max_nz, axis=0)[self.max_nz:]
+        # Set these entries to -inf.
+        X[non_top_indices, np.arange(X.shape[1])] = -np.inf
+        max_X = np.maximum(X, 0)
+        val = np.sum(max_X ** 2, axis=0) / (2 * self.gamma)
+        G = max_X / self.gamma
+        return val, G
+
+    def max_Omega(self, X, b):
+        # For each column of X, find top max_nz values and
+        # their corresponding indices.
+        max_nz_indices = np.argpartition(
+            X,
+            kth=-self.max_nz,
+            axis=0)[-self.max_nz:]
+        max_nz_values = X[max_nz_indices, np.arange(X.shape[1])]
+
+        # Project the top max_nz values onto the simplex.
+        G_nz_values = ot.smooth.projection_simplex(
+            max_nz_values / (b * self.gamma), axis=0)
+
+        # Put the projection of max_nz_values to their original indices
+        # and set all other values zero.
+        G = np.zeros_like(X)
+        G[max_nz_indices, np.arange(X.shape[1])] = G_nz_values
+        val = np.sum(X * G, axis=0)
+        val -= 0.5 * self.gamma * b * np.sum(G * G, axis=0)
+        return val, G
+
+    def Omega(self, T):
+        return 0.5 * self.gamma * np.sum(T ** 2)
+
+
+def sparsity_constrained_ot_dual(
+        a, b, M, reg, max_nz,
+        method="L-BFGS-B", stopThr=1e-9,
+        numItermax=500, verbose=False, log=False):
+    r"""
+    Solve the sparsity-constrained OT problem in the dual and return the OT matrix.
+
+    The function solves the sparsity-contrained OT in dual formulation in
+    :ref:`[50] <references-sparsity-constrained-ot-dual>`.
+
+
+    Parameters
+    ----------
+    a : np.ndarray (ns,)
+        samples weights in the source domain
+    b : np.ndarray (nt,) or np.ndarray (nt,nbb)
+        samples in the target domain, compute sinkhorn with multiple targets
+        and fixed :math:`\mathbf{M}` if :math:`\mathbf{b}` is a matrix
+        (return OT loss + dual variables in log)
+    M : np.ndarray (ns,nt)
+        loss matrix
+    reg : float
+        Regularization term >0
+    max_nz: int
+        Maximum number of non-zero entries permitted in each column of the
+        optimal transport matrix.
+    method : str
+        Solver to use for scipy.optimize.minimize
+    numItermax : int, optional
+        Max number of iterations
+    stopThr : float, optional
+        Stop threshold on error (>0)
+    verbose : bool, optional
+        Print information along iterations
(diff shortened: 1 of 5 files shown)


liutianlin0121:
Question: In the `ot/smooth.py` module, it seems that only the dual formulation `smooth_ot_dual` supports multiple backends while the semi-dual formulation `smooth_ot_semi_dual` does not. I follow this convention in my pull request---only the dual formulation in `ot/sparse.py` calls `get_backend`. However, is there an obstacle to making the semi-dual formulation compatible with different backends too?

ot/sparse.py
```diff
@@ -0,0 +1,229 @@
+"""
+Sparsity-constrained optimal transport solvers.
+
+Implementation of :
+Sparsity-Constrained Optimal Transport.
+Tianlin Liu, Joan Puigcerver, Mathieu Blondel.
+In Proc. of AISTATS 2018.
+https://arxiv.org/abs/1710.06276
```
mblondel, line 8: Wrong ref and link

test/test_sparse.py
```diff
@@ -0,0 +1,62 @@
+"""Tests for ot.smooth model """
```
mblondel, line 1: ot.sparse

ot/sparse.py
```diff
@@ -0,0 +1,229 @@
...
+        # For each column of X, find top max_nz values and
+        # their corresponding indices.
+        max_nz_indices = np.argpartition(
+            X,
+            kth=-self.max_nz,
+            axis=0)[-self.max_nz:]
+        max_nz_values = X[max_nz_indices, np.arange(X.shape[1])]
+
+        # Project the top max_nz values onto the simplex.
+        G_nz_values = ot.smooth.projection_simplex(
```
mblondel, line 54: projection_simplex is also doing a sort, so you end up doing two sorts. You could create a function `projection_sparse_simplex` that does only one sort instead (potentially leave it as a todo for the time being)

test/test_sparse.py
liutianlin0121: done

liutianlin0121:
Thanks! I updated the pull request.

rflamary:
Hello thanks for the PR, we will do a code reveiw ASAP!

ot/sparse.py
```diff
@@ -0,0 +1,279 @@
...
+    def delta_Omega(self, X):
+        # For each column of X, find entries that are not among the top max_nz.
+        non_top_indices = np.argpartition(
+            -X, self.max_nz, axis=0)[self.max_nz:]
+        # Set these entries to -inf.
+        X[non_top_indices, np.arange(X.shape[1])] = -np.inf
+        max_X = np.maximum(X, 0)
+        val = np.sum(max_X ** 2, axis=0) / (2 * self.gamma)
+        G = max_X / self.gamma
+        return val, G
```
mblondel, line 105: Using [check_grad](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.check_grad.html), you can write a unit test checking that `G` is the correct gradient for `val`.

ot/sparse.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def max_Omega(self, X, b):
+        # Project the scaled X onto the simplex with sparsity constraint.
+        G = projection_sparse_simplex(
+            X / (b * self.gamma), self.max_nz, axis=0)
+        val = np.sum(X * G, axis=0)
+        val -= 0.5 * self.gamma * b * np.sum(G * G, axis=0)
+        return val, G
```
mblondel, line 113: Same here.

mblondel (commented):
Thanks Tian Lin! Two other comments below.

rflamary:
Also small comment, since you are using functions from `ot.smooth` and your paper is clearly related I would put the new solvers in` ot.smooth` and the function  `projection_sparse_simplex` in ot.utils (we might need it somewhere else ;) ).

liutianlin0121:
Hey both, thanks! I incorporated the suggestions. 

Note that I only added the `check_grad` test for the `SparsityConstrained` regularization. Similar grad checks can be added to other regularization classes, like `NegEntropy` and `SquaredL2`. But to do that for `SquaredL2`, I think I'll need to modify the function `projection_simplex` so that it can accept 1-dim array. The reason is that the objective function used by `check_grad` has to be a scalar-valued one. This means that we need to supply a one-dim array `X` as the input of `max_Omega`, which currently raises an error because `projection_simplex` only works for 2-dim arrays. If you think it is necessary, I can modify `projection_simplex` in a way similar to `projection_sparse_simplex` and then add gradient checks on `SquaredL2`.

ot/smooth.py
```diff
@@ -636,3 +683,166 @@ def smooth_ot_semi_dual(a, b, M, reg, reg_type='l2', method="L-BFGS-B", stopThr=
         return G, log
     else:
         return G
+
+
+def sparsity_constrained_ot_dual(
```
mblondel, line 688: You could reuse the existing functions with `reg_type="sparse_l2"` or `reg_type="l2_topk"` instead. You would need to add `max_nz` but I think that's fine.

examples/plot_OT_1D_smooth.py
```diff
@@ -110,3 +110,15 @@
 ot.plot.plot1D_mat(a, b, Gsm, 'OT matrix Smooth OT l2 reg.')
 
 pl.show()
+
+#%% Sparsity-constrained OT
```
rflamary, line 114: you cold rename the exmaple "Smooth and sparse OT" and maybe remove sinkhorn and KL to fous on the new solvers more?

ot/smooth.py
```diff
@@ -26,25 +26,40 @@
 #         Remi Flamary <remi.flamary@unice.fr>
```
rflamary, line 26: add your name here

ot/smooth.py
```diff
@@ -26,25 +26,40 @@
...
 
 """
-Smooth and Sparse Optimal Transport solvers (KL an L2 reg.)
+Smooth and Sparse Optimal Transport solvers (KL an L2 reg.) [17] and sparsity
```
rflamary, line 29: Use a shorter first line without refs for here:
image

Also cleanup the text for the module that is not looking good:
image

ot/smooth.py
```diff
@@ -435,7 +483,8 @@ def get_plan_from_semi_dual(alpha, b, C, regul):
     return regul.max_Omega(X, b)[1] * b
 
 
-def smooth_ot_dual(a, b, M, reg, reg_type='l2', method="L-BFGS-B", stopThr=1e-9,
+def smooth_ot_dual(a, b, M, reg, reg_type='l2', max_nz=None,
```
rflamary, line 486: add `max_nz`  at the end in order to not berak the function call

ot/utils.py
```diff
@@ -117,6 +117,74 @@ def proj_simplex(v, z=1):
...
+
+    References:
+        Sparse projections onto the simplex
+        Anastasios Kyrillidis, Stephen Becker, Volkan Cevher and, Christoph Koch
+        ICML 2013
+        https://arxiv.org/abs/1206.1529
+    """
+
+    if V.ndim == 1:
+        return projection_sparse_simplex(
```
rflamary, line 148: love this line, `projection_simplex` above should have it also instead of the crappy d1 thingy above

ot/utils.py
```diff
@@ -117,6 +117,74 @@ def proj_simplex(v, z=1):
...
+            V[np.newaxis, :], max_nz, z, axis=1).ravel()
+
+    if V.ndim > 2:
+        raise ValueError('V.ndim must be <= 2')
+
+    if axis == 1:
```
rflamary, line 154: could you pelase do a backend compatible version of the function (see above)? Not that the backend could be passed optionally to the function to avoid the test zand initialized as `nx=None`. This is not necessary but it would means you ca use your projection in torch and differentiate  (what can be) the output.

rflamary (commented):
hello @liutianlin0121 ,

Thanks so much for the PR. I still have a few comments before merging but we are neraly there.

rflamary (commented):
LGTM. @liutianlin0121 could you please add you PR to the RELEASES.md file with a short description of the new feature? I thank after that we can merge.

PR #468, opened 2023-05-02
Improve Bures-Wasserstein distance

francois-rozet:
## Types of changes

* Fixed typo in the documentation of the Bures-Wasserstein distance ($\Sigma_s$ instead of $\Sigma_s^{1/2}$).
* Faster way of computing the trace of the square-root of the product of $\Sigma_s$ and $\Sigma_t$.

The implementation is based on two facts:

1. The trace of $A$ equals the sum of its eigenvalues.
2. The eigenvalues of $\sqrt{A}$ are the square-roots of the eigenvalues of $A$.

Then, $\mathrm{tr}(\sqrt{A})$ is the sum of the square-roots of the eigenvalues of $A$.

See https://github.com/Lightning-AI/torchmetrics/issues/1705.

Computing the square-root of a matrix is slow and unstable.

The new implementation still passes the tests (at least with NumPy backend).

## PR checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,7 @@
 - Fix issues with cuda for ot.binary_search_circle and with gradients for ot.sliced_wasserstein_sphere (PR #457)
 - Major documentation cleanup (PR #462, #467)
 - Fix gradients for "Wasserstein2 Minibatch GAN" example (PR #466)
+- Faster Bures-Wasserstein distance with NumPy backend (PR #468)
 
 ## 0.9.0
 
--- a/ot/backend.py
+++ b/ot/backend.py
@@ -844,6 +844,16 @@ class Backend():
         """
         raise NotImplementedError()
 
+    def eigvals(self, a):
+        r"""
+        Computes the matrix eigenvalues. Requires input to be square.
+
+        This function follows the api from :any:`np.linalg.sqrtm`.
+
+        See: https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigvals.html
+        """
+        raise NotImplementedError()
+
     def sqrtm(self, a):
         r"""
         Computes the matrix square root. Requires input to be definite positive.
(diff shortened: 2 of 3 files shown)


rflamary:
Hello @francois-rozet, when a sepcific backend type misses a feature is a feature we usually add it to the backend, in this case maybe we shoudl add `nx.real` and `nx.imag`to the backend functions

rflamary:
Also @francois-rozet  could you please do a quick test of the timing before and after your speedup for different backends (at least numpy and pytorch)  and put it in the text description of the PR?  

I like to have quantified performance gain in the history/github to know why we changed stuff. Also maybe a quick test that checks that the new function returns the same thing as the np.trace () up to numerical precision. It seems right but such a test will help detecting potential problems in the future.

francois-rozet:
After some tests, I found that this implementation is only faster for NumPy. Computing the square root of a general matrix is indeed slower than computing its eigenvalues. However, computing the square root of a symmetric matrix takes more or less the same time as computing its eigenvalues. In fact, the PyTorch backend uses `torch.linalg.eigh` to implement `sqrtm`. So I think instead of modifying the algorithm, we can simply replace the `sqrtm` of the NumPy backend.

rflamary:
OK it make sens, happy I pushed you to investigate. You can leave the envals function in the backend it can be usefull in the future (trace norm regularization for instance)

francois-rozet:
I added a line to the `RELEASES.md` file. Also, here is a small Jupyter notebook demo to show that `np.linalg.eigh` is much faster than `scipy.linalg.sqrtm`:

```python
import numpy as np
import scipy.linalg as sl

A = np.random.rand(512, 512) / 512 ** 0.5
A = A @ A.T + np.eye(512) * 1e-6

%timeit np.linalg.eigh(A)
%timeit sl.sqrtm(A)
```

returns

```
19.5 ms ± 2.27 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
166 ms ± 15.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Issue #465, opened 2023-04-28
The function 'cost_normalization' do not use the backend and only use numpy


framunoz:
## Describe the bug
I was using the `ot.da.SinkhornTransport` class when I got the following warning:
```UserWarning: Warning: numerical errors at iteration 0```.

Thinking about why I got this error, I remembered that cost matrices should be normalised, so I changed the `norm` parameter to `"max"` to normalise it to the maximum value. However, this gave me an error.

The reason was that the `utils.cost_normalization` function is implemented only using numpy, when I'm currently using pytorch tensors. Internally, I got the backend of the `C` array that uses this function, and changed the line `C /= float(np.max(C))` to `C /= float(nx.max(C))`, and then the program worked fine.

I mention this mainly because it may be a bug that happens to someone else in another context.

### Environment (please complete the following information):
- OS (e.g. MacOS, Windows, Linux): Window
- Python version: 3.10
- How was POT installed (source, `pip`, `conda`): pip
- Build command you used (if compiling from source):
- Only for GPU related bugs:
    - CUDA version:
    - GPU models and configuration:
    - Any other relevant information:

Output of the following code snippet:
```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import ot; print("POT", ot.__version__)
```

```
Windows-10-10.0.19045-SP0
Python 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
NumPy 1.24.1
SciPy 1.9.3
POT 0.8.2
```

rflamary:
Good point and your fix seems perfect, those are the edge cases that were not detected when converting to backend.

Could you do this short PR for us please (and add a test that check that normalization does not breaks with non-numpy arrays)?

framunoz:
Hi,

I was writing the code to try to fix this bug in the branch "`cost_normalization_debug`". In order to fix this bug, I had to create the `median` method in the backend class, and implement this method in the rest of the backends. However, I realised that tensorflow doesn't have a `median` function, although it may be replicable with the percentile function (as seen in this Stack Overflow question [Tensorflow median value](https://stackoverflow.com/questions/43824665/tensorflow-median-value)). Also, I couldn't run the test scripts, as I didn't get the Jax library installed correctly.

[...]

PR #472, opened 2023-05-03
Refactor the function `utils.cost_normalization` to work with multiple backends

framunoz:
## Types of changes

* The `utils.cost_normalization` function now uses backends in a general way.
* The  backends now implements the 'median' function, with exception of tensorflow backend.

The motivation for making this change is that currently, when using a backend other than Numpy (e.g. PyTorch), the program crashes when trying to use the `utils.cost_normalization` function. This function is also essential in the `ot.da.SinkhornTransport` class, for example.

This PR fix the issue #465

I modified the existing test `test_cost_normalization` to use the different types of backends, and then passed them back to Numpy again to keep the tests that existed before.

In addition, I added new cases to the backend tests, to include the new `median` function.

## PR checklist

- [ ] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.

### DIFF
--- a/ot/backend.py
+++ b/ot/backend.py
@@ -574,6 +574,16 @@ class Backend():
         """
         raise NotImplementedError()
 
+    def median(self, a, axis=None):
+        r"""
+        Computes the median of a tensor along given dimensions.
+
+        This function follows the api from :any:`numpy.median`
+
+        See: https://numpy.org/doc/stable/reference/generated/numpy.median.html
+        """
+        raise NotImplementedError()
+
     def std(self, a, axis=None):
         r"""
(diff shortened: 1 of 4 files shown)


rflamary:
Hello this is a pain for tensorflow, You shouldl the numpy median and then convert back to tensorflow afterward in the backend. You loose differentiability but median is non differentiable anyways and it is not necessary for DA methods and  it will work. maybe just do a warning with wrarning.warning that median in computed using numpy and the array is detached in the tf backdn function.

framunoz:
I see that the error raised in the test was due to an incompatibility of the version with Pytorch. Maybe with a similar strategy used in Tensorflow will work?

PR #505, opened 2023-08-18
[MRG] Gracefully handle cost = 0 in `generic_conditional_gradient`

kachayev:
## Types of changes

`relative_delta_cost_G` in `generic_conditional_gradient` is only computed when the cost != 0.

As the contract for cost evaluation in `generic_conditional_gradient` is rather generic, it could be the case it's set to 0. In this case computing `relative_delta_cost_G` fails because of zero division. As we only need the value of `relative_delta_cost_G` for early stopping condition, it doesn't hurt to skip the computation when cost is 0.

The code to reproduce the issue:

```python
import numpy as np
import ot
import ot.datasets

n_samples = 50
mu_s = np.array([0, 0])
cov_s = np.array([[1, 0], [0, 1]])
xs = ot.datasets.make_2D_samples_gauss(n_samples, mu_s, cov_s, random_state=4)
xt = xs[::-1].copy()
p = ot.unif(n_samples)
q = ot.unif(n_samples)
G0 = p[:, None] * q[None, :]
C1 = ot.dist(xs, xs)
C2 = ot.dist(xt, xt)
C1 /= C1.max()
C2 /= C2.max()
ot.gromov.gromov_wasserstein(C1, C2, p, q, 'square_loss', G0=G0, verbose=True)
```

This actually happens in `test_gromov_dtype_device` test case. The test case is not modified to ensure we don't have warnings when computing `ot.gromov.gromov_wasserstein`.

## PR checklist


cedricvincentcuaz:
Hello @kachayev,
Sorry for the late answer. In the end, these indeed seem to be really specific edge cases, I am not so sure that it is worth it to add dedicated tests to gain one iteration when these occur.
I suggest to keep the previous behaviour as your code currently does. So we could merge after tests are completed.

Overall, I think that it would be nice to go a normalisation of stopping criterions for various solvers (at least two consistent types) in POT. So our current interrogations could be handled in this context.

Best, Cédric.

PR #507, opened 2023-08-20
Vectorize cost and gradient for `ot.da.sinkhorn_l1l2_gl`

kachayev:
## Types of changes

Computation of the cost and gradient callbacks for generalized conditional gradient used by `ot.da.sinkhorn_l1l2_gl` are now vectorized.

One potential issue here is that `norm` function for backends were implemented as a direct computation of Euclidean 2-norm without leveraging built-in `norm`s. I don't know if that was done on purpose (maybe there's a known problem with the one exposed by each library). I switched the call to make access to additional arguments, like `axis` and `keepdims`. Still could be implemented manually here, if necessary.

Overall sequential code is much harder to optimize compared to vectorized execution, specifically when working with GPUs. In this particular case, it's also easy to spot that `labels_a == lab` indices [here](https://github.com/PythonOT/POT/blob/9dd4d8dd987d2b70284cce3a61793afcc5903902/ot/da.py#L244) are recomputed within nested for-loop per each call of the function (n_samples * n_labels * n_calls) despite being static. New implementation prepares matrix with proper indexing (only once) and then computes the result by leveraging normalization/summation operations on matrices defined by the corresponding backend.

Also, it seems like now we can have this function working normally on JAX backend (previous version didn't work because of immutability of JAX tensors).

New unit test `test_sinkhorn_l1l2_gl_cost_vectorized` is defined for `test_da.py` test suite. This might be not the best way to do a unit testing but for this specific use case seems fair: it just contains 2 implementations (old and new one) to make sure those function return the same result.

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![PyPI version](https://badge.fury.io/py/POT.svg)](https://badge.fury.io/py/POT)
 [![Anaconda Cloud](https://anaconda.org/conda-forge/pot/badges/version.svg)](https://anaconda.org/conda-forge/pot)
-[![Build Status]([https://github.com/PythonOT/POT/workflows/build/badge.svg?branch=master&event=push](https://github.com/PythonOT/POT/actions/workflows/build_tests.yml/badge.svg))](https://github.com/PythonOT/POT/actions)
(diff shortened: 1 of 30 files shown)


kachayev:
Tests are okay, finally.

PR #510, opened 2023-08-23
Use local random state in `ot.datasets.make_data_classif`

kachayev:
## Types of changes

Switch from `np.random.randn` to `rng.randn` when generating `make_data_classif` dataset with `dataset='2gauss_prop'`.

## Motivation and context / Related issue

It's not desirable to change global state. And it's already done for other datasets in the same function.

## How has this been tested (if it applies)

n/a

## PR checklist

- [x] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [x] The documentation is up-to-date with the changes I made (check build artifacts).
- [x] All tests passed, and additional code has been **covered with new tests**.
- [ ] I have added the PR and Issue fix to the [**RELEASES.md**](RELEASES.md) file.

PR #512, opened 2023-08-24
Overhaul random state management

kachayev:
## Types of changes

This is rather lengthy follow-up for #510. Hopefully I found all cases to update:
* manual seeding is removed from tests
* tests are now using local random state generators rather than global state
* functions that rely on randomization except `random_state` arguments now

To make reproducibility easier.

Test cases are updated accordingly. It's interesting that previously for some of them (specifically suites from test_gromov) required explicit seeding for tests to work. Now they rely on `random_state` passed into functions and don't break because of global state.

## PR checklist

I guess new parameters added to the API should be mentioned in the RELEASE file.

### DIFF
--- a/ot/dr.py
+++ b/ot/dr.py
@@ -25,7 +25,7 @@ import pymanopt.manifolds
 import pymanopt.optimizers
 
 from .bregman import sinkhorn as sinkhorn_bregman
-from .utils import dist as dist_utils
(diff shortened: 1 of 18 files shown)


agramfort (approved):

agramfort:
thx @kachayev !

PR #533, opened 2023-10-13
[MRG] Debug convolutional methods that compute barycenters to work with different devices.

framunoz:
## Types of changes

* [X] The `linspace` method of the backends now has the `type_as` argument to convert to the same dtype and device.
* [X] The `convolutional_barycenter2d` and `convolutional_barycenter2d_debiased` functions now work with different devices.

The problem that the PR solves is that the `convolutional_barycenter2d` and `convolutional_barycenter2d` functions were not working on different devices (e.g. using PyTorch with CUDA). It was detected that the error came from the `linspace` method of the backends not having a `type_as` argument.

* Add a test in `test_backend.py` that checks if the `type_as` argument in the `linspace` method works.
* Add the tests `test_wasserstein_bary_2d_d_dtype_device`, `test_wasserstein_bary_2d_device_tf`, `test_wasserstein_bary_2d_debiased_dtype_device` and `test_wasserstein_bary_2d_debiased_device_tf` which mirror the `test_sinkhorn2_variants_dtype_device` and `test_sinkhorn2_variants_device_tf` tests to check if the barycenter work with different dtypes and devices.

## PR checklist

- [X] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.
- [X] The documentation is up-to-date with the changes I made (check build artifacts).
- [X] All tests passed, and additional code has been **covered with new tests**.

### DIFF
--- a/test/test_bregman.py
+++ b/test/test_bregman.py
@@ -945,7 +945,7 @@ def test_wasserstein_bary_2d_debiased_dtype_device(nx, method):
 def test_wasserstein_bary_2d_debiased_device_tf(method):
     # Using the Tensorflow backend
     nx = ot.backend.TensorflowBackend()
-    
+
     rng = np.random.RandomState(42)
     size = 20  # size of a square image
 
@@ -1018,7 +1018,6 @@ def test_wasserstein_bary_2d_debiased_device_tf(method):
             # Test that the dtype and device are the same after the computation
             nx.assert_same_dtype_device(Ab, bary_wass_b)
             assert nx.dtype_device(bary_wass_b)[1].startswith("GPU")
-
 
 
 def test_unmix(nx):
--- a/test/test_bregman.py
+++ b/test/test_bregman.py
@@ -995,7 +995,6 @@ def test_wasserstein_bary_2d_debiased_device_tf(method):
             # Test that the dtype and device are the same after the computation
             nx.assert_same_dtype_device(Ab, bary_wass_b)
 
-    
     # Check that everything happens on the GPU
     Ab = nx.from_numpy(A)
 
--- a/test/test_bregman.py
+++ b/test/test_bregman.py
@@ -31,23 +31,19 @@ def test_sinkhorn(verbose, warn):
     G = ot.sinkhorn(u, u, M, 1, stopThr=1e-10, verbose=verbose, warn=warn)
 
     # check constraints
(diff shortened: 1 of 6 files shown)


test/test_bregman.py
```diff
@@ -352,13 +352,15 @@ def test_sinkhorn2_variants_device_tf(method):
         nx.assert_same_dtype_device(Mb, Gb)
         nx.assert_same_dtype_device(Mb, lossb)
 
+    # Check that everything happens on the GPU
```
rflamary, line 355: why did you move this outside of the GPU test?

test/test_bregman.py
```diff
@@ -726,6 +728,137 @@ def test_wasserstein_bary_2d(nx, method):
...
+            nx.assert_same_dtype_device(Ab, bary_wass_b)
+
+
+@pytest.mark.skipif(not tf, reason="tf not installed")
+@pytest.mark.parametrize("method", ["sinkhorn", "sinkhorn_log"])
+def test_wasserstein_bary_2d_device_tf(method):
+    # Using the Tensorflow backend
+    nx = ot.backend.TensorflowBackend()
+
+    rng = np.random.RandomState(42)
```
rflamary, line 786: could you factorize all this data initialization in a function ad call it each time please (there are other plces where it will be used)?

rflamary (commented):
Thanks @framunoz  for you PR, it looks great. Could you answer my question and comment  please?

test/test_bregman.py
framunoz: Because Codecov does not test these lines, since it works on cpu, giving less coverage to the overall code. When I leave the device verification to the end, Codecov will give coverage to all lines except the last one. Otherwise, it will not give coverage to most of the lines after the `if` as it was before.

To summarize, it is rather a trick to increase the coverage, which were cases that can still be included, but due to limitations of the Codecov machine, it gives the impression that it worsens the coverage.

test/test_bregman.py
framunoz: I included a function that initializes the matrix A, and replaced it each time it is called.

framunoz:
The extension that reformats the code to comply with the PEP 8 format made many changes to the code. Do I leave it as it was before the commit? Or is the new style preferred?

framunoz:
I used black in the tests, although because of the vscode extension, the formatting was applied automatically.

In these last commits I left the formatting of the rest of the tests as they were before, also, I added a new line in the `.gitignore` to ignore the `.venv`.

PR #539, opened 2023-10-25
[MRG] Extend unbalanced OT solvers for more flexibility

6Ulm:
## Types of changes

- Modify all unbalanced solvers in `unbalanced.py ` (except for barycenter), so that the argument `reg_m` can take either a scalar, or an indexable object of length 1 or 2. 
- Edit `mm_unbalanced` method so that it can be used for both unregularized and regularized problems (previously supported unregularized only). Also, fix an implementation error of matrix $K$ in the $L_2$ case.
- Fix the KL formula and the corresponding gradient for L-BFGS-B solver.
- Add corresponding tests to `test_unbalanced.py`.

1. The current solvers for unbalanced OT have some limitations, which may restrict their usage in practice.
- The `sinkhorn_unbalanced` and `lbfgsb_unbalanced` methods only allow for `reg_m` to be a scalar, thus impose the same penalization on both marginals.
- The `mm_unbalanced` method has the same limitation and restrict only to the unregularized problem.
2.  The `_get_loss_unbalanced` method uses incorrect formula of KL divergence, since we are working with unnormalized measures. As a result, the calculation of the gradient is also incorrect.

This PR fixes all of these issues. Moreover, as a byproduct, the new version of `sinkhorn_unbalanced` method also allows to solve the entropic balanced and semi-relaxed problems.

Tested when `reg_m` is a scalar, indexable objects of length 1 and 2 (tuple, list, array/tensor).

## PR checklist

### DIFF
--- a/ot/unbalanced.py
+++ b/ot/unbalanced.py
@@ -1151,11 +1151,11 @@ def mm_unbalanced(a, b, M, reg_m, reg=0, div='kl', G0=None, numItermax=1000,
     >>> b=[.5, .5]
     >>> M=[[1., 36.],[9., 4.]]
     >>> np.round(ot.unbalanced.mm_unbalanced(a, b, M, 5, div='kl'), 2)
-    array([[0.45 , 0.  ],
-           [0.   , 0.34]])
+    array([[0.45, 0.  ],
+           [0.  , 0.34]])
     >>> np.round(ot.unbalanced.mm_unbalanced(a, b, M, 5, div='l2'), 2)
-    array([[0.4 , 0. ],
-           [0.  , 0.1]])
+    array([[0.4, 0. ],
+           [0. , 0.1]])
 
 
     .. _references-regpath:
@@ -1475,11 +1475,11 @@ def lbfgsb_unbalanced(a, b, M, reg, reg_m, reg_div='kl', regm_div='kl', G0=None,
     >>> b=[.5, .5]
     >>> M=[[1., 36.],[9., 4.]]
     >>> np.round(ot.unbalanced.lbfgsb_unbalanced(a, b, M, reg=0, reg_m=5, reg_div='kl', regm_div='kl'), 2)
-    array([[0.45 , 0.  ],
-           [0.   , 0.34]])
+    array([[0.45, 0.  ],
+           [0.  , 0.34]])
     >>> np.round(ot.unbalanced.lbfgsb_unbalanced(a, b, M, reg=0, reg_m=5, reg_div='l2', regm_div='l2'), 2)
-    array([[0.4 , 0. ],
-           [0.  , 0.1]])
+    array([[0.4, 0. ],
+           [0. , 0.1]])
 
     References
     ----------
--- a/ot/unbalanced.py
+++ b/ot/unbalanced.py
@@ -1319,7 +1319,7 @@ def mm_unbalanced2(a, b, M, reg_m, reg=0, div='kl', G0=None, numItermax=1000,
     ot.lp.emd2 : Unregularized OT loss
     ot.unbalanced.sinkhorn_unbalanced2 : Entropic regularized OT loss
     """
-    _, log_mm = mm_unbalanced(a, b, M, reg, reg_m, div=div, G0=G0,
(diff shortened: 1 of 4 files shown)


ot/unbalanced.py
```diff
@@ -15,10 +17,9 @@
 
 from .backend import get_backend
 from .utils import list_to_array
-# from .utils import unif, dist
 
 
-def sinkhorn_unbalanced(a, b, M, reg, reg_m, method='sinkhorn', numItermax=1000,
+def sinkhorn_unbalanced(a, b, M, reg, reg_m=None, method='sinkhorn', numItermax=1000,
```
rflamary, line 22: sinkhorn unbalanced should always requite reg_m it is in thje name. abut indeed do the balanc"d one for reg_m=[inf,inf]

ot/unbalanced.py
```diff
@@ -147,7 +154,7 @@ def sinkhorn_unbalanced(a, b, M, reg, reg_m, method='sinkhorn', numItermax=1000,
         raise ValueError("Unknown method '%s'." % method)
 
 
-def sinkhorn_unbalanced2(a, b, M, reg, reg_m, method='sinkhorn',
+def sinkhorn_unbalanced2(a, b, M, reg, reg_m=None, method='sinkhorn',
```
rflamary, line 157: same here

ot/unbalanced.py
```diff
@@ -376,6 +394,16 @@ def sinkhorn_knopp_unbalanced(a, b, M, reg, reg_m, numItermax=1000,
     else:
         n_hists = 0
 
+    if isinstance(reg_m, float) or isinstance(reg_m, int):
```
rflamary, line 397: or len(reg_m) == 1 ?

if it us a numpy or torch array of length 1 it shoud also work, please check

ot/unbalanced.py
```diff
@@ -559,6 +594,16 @@ def sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, tau=1e5, numItermax=1000
...
```
rflamary, line 597: same comment

ot/unbalanced.py
```diff
@@ -559,6 +594,16 @@ def sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, tau=1e5, numItermax=1000
...
+        reg_m1, reg_m2 = reg_m, reg_m
+    elif reg_m is None:
+        reg_m1, reg_m2 = float("inf"), float("inf")
```
rflamary, line 600: please upadet test to handle all cases to check that nothing breaks

ot/unbalanced.py
```diff
@@ -1131,30 +1185,41 @@ def mm_unbalanced(a, b, M, reg_m, div='kl', G0=None, numItermax=1000,
...
+            raise ValueError("Epsilon must be either a scalar or an indexable object of length 2.")
+        else:
+            reg_m1, reg_m2 = reg_m[0], reg_m[1]
+
     if log:
         log = {'err': [], 'G': []}
 
     if div == 'kl':
-        K = nx.exp(M / - reg_m / 2)
+        sum_r = reg + reg_m1 + reg_m2
```
rflamary, line 1200: those shoul be done outside of the loop non? smae with r1

ot/unbalanced.py
```diff
@@ -1131,30 +1185,41 @@ def mm_unbalanced(a, b, M, reg_m, div='kl', G0=None, numItermax=1000,
...
-        K = nx.maximum(a[:, None] + b[None, :] - M / reg_m / 2,
-                       nx.zeros((dim_a, dim_b), type_as=M))
+        K = reg_m1 * a[:, None] + reg_m2 * b[None, :] + reg * a[:, None] * b[None, :] - M
+        K = nx.maximum(K, nx.zeros((dim_a, dim_b), type_as=M))
     else:
         warnings.warn("The div parameter should be either equal to 'kl' or \
                       'l2': it has been set to 'kl'.")
         div = 'kl'
-        K = nx.exp(M / - reg_m / 2)
+        sum_r = reg + reg_m1 + reg_m2
```
rflamary, line 1210: please add a test to handle all thso cases

rflamary (commented):
Thanks Huy for the big update that was needed. I have a few comments but it is looking good.

PR #544, opened 2023-10-26
[MRG] Add LazyTensor for large scale OT

rflamary:
## Types of changes

This PR adds a very simple LazyTensor class that can implement large scale matrices and tensors that are computed only lazily when getting slices of the tensor.

This class will be very interesting when implementing large scale OT problems such as factored OT, low rank OT and lazy implementation of sinkhorn. The objective is to be very simple with a storage of all necessary data and a function that compute the values on sliced (without computing the full tensor).

Exemple of use

```python
import numpy as np
from ot.utils import LazyTensor, 

n1=100
n2=200
x1 = np.random.randn(n1,2)
x2 = np.random.randn(n2,2)

# i,j can be integers or slices, x1,x2 have to be passed as keyword arguments
def getitem(i,j, x1, x2):
    return np.dot(x1[i], x2[j].T)

# create a lazy tensor (giv data as arguments)
T = LazyTensor((n1,n2),getitem, x1=x1, x2=x2)

print(T.shape)
# (100, 200)
print(T)
# LazyTensor(shape=(100, 200),attributes=(x1,x2))

# get the full tensor (not lazy)
full_T = T[:]

# get one component
T11 = T[1,1]

# get one row
T1 = T[1]

# get one column with slices
Tsliced = T[::10,5]

# get the data as attributes
x1_0 = T.x1
x2_0 = T.x2

# create a LazyTensor from another one : exp(T)
T2 = ot.utils.LazyTensor(T.shape,lambda i,j,T: np.exp(T[i,j]) , T=T)

# compute sum
s = reduce_lazytensor(T, np.sum) 

# compute marginals
c = reduce_lazytensor(T, np.sum, axis=1) 
r = reduce_lazytensor(T, np.sum, axis=0, batch_size=50) 

```

## PR checklist

### DIFF
--- a/ot/utils.py
+++ b/ot/utils.py
@@ -491,7 +491,7 @@ def get_coordinate_circle(x):
     x_t = (nx.atan2(-x[:, 1], -x[:, 0]) + np.pi) / (2 * np.pi)
     return x_t
 
-  
+
 def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
     """ Reduce a LazyTensor along an axis with function fun using batches.
 
@@ -524,7 +524,7 @@ def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
     """
 
     if nx is None:
-        nx = get_backend(a)
+        nx = get_backend(a[0])
 
     if batch_size is None:
         batch_size = 100
@@ -551,13 +551,11 @@ def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
         else:
             shape = (a.shape[0], *a.shape[2:])
         res = nx.zeros(shape, type_as=a[0])
-        if nx.__name__ == "jax":
+        if nx.__name__ in ["jax", "tf"]:
             lst = []
             for i in range(0, a.shape[0], batch_size):
                 lst.append(fun(a[i:i + batch_size], 1))
             return nx.concatenate(lst, axis=0)
-        elif nx.__name__ == "tf":
-            raise (NotImplementedError("LazyTensor reduction not implemented for TF backend."))
         else:
             for i in range(0, a.shape[0], batch_size):
                 res[i:i + batch_size] = fun(a[i:i + batch_size], axis=1)
@@ -565,6 +563,7 @@ def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
 
     else:
         raise (NotImplementedError("Only axis=None is implemented for now."))
+
 
 def get_parameter_pair(parameter):
     r"""Extract a pair of parameters from a given parameter
--- a/ot/utils.py
+++ b/ot/utils.py
@@ -562,7 +562,7 @@ def reduce_lazytensor(a, func, axis=None, nx=None, batch_size=100):
         return res
 
     else:
-        raise (NotImplementedError("Only axis=None is implemented for now."))
+        raise (NotImplementedError("Only axis=None, 0 or 1 is implemented for now."))
 
 
 def get_lowrank_lazytensor(Q, R, d=None, nx=None):
--- a/ot/utils.py
+++ b/ot/utils.py
@@ -492,18 +492,22 @@ def get_coordinate_circle(x):
     return x_t
 
 
-def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
+def reduce_lazytensor(a, func, axis=None, nx=None, batch_size=None):
     """ Reduce a LazyTensor along an axis with function fun using batches.
 
     When axis=None, reduce the LazyTensor to a scalar as a sum of fun over
     batches taken along dim.
 
+    .. warning:: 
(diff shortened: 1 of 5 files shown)


test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+    T[:]
+
+    # get one component
+    T[1, 1]
+
+    # get one row
+    T[1]
+
+    # get one column with slices
+    T[::10, 5]
```
agramfort, line 441: can you add that the outputs are correct?

test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+
+
+def test_OTResult_LazyTensor(nx):
+
+    n1 = 100
+    n2 = 200
+
+    rng = np.random.RandomState(42)
+    a = rng.rand(n1)
+    a = a / a.sum()
```
agramfort, line 451:
```suggestion
    a /= a.sum()
```

test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+    b = rng.rand(n2)
+    b = b / b.sum()
```
agramfort, line 453:
```suggestion
    b /= b.sum()
```

test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+
+
+def test_LazyTensor_reduce(nx):
+
+    n1 = 100
+    n2 = 200
+
+    rng = np.random.RandomState(42)
+    a = rng.rand(n1)
+    a = a / a.sum()
```
agramfort, line 476: /=

test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+    b = rng.rand(n2)
+    b = b / b.sum()
```
agramfort, line 478: /=

test/test_utils.py
```diff
@@ -401,3 +402,111 @@ def test_get_coordinate_circle():
...
+
+    a, b = nx.from_numpy(a, b)
+
+    def getitem(i, j, a, b):
+        return a[i, None] * b[None, j]
+
+    # create a lazy tensor
+    T = ot.utils.LazyTensor((n1, n2), getitem, a=a, b=b)
```
agramfort, line 486: I would create fixture for T to avoid duplicating T creation in all your tests

ot/utils.py
```diff
@@ -492,6 +492,79 @@ def get_coordinate_circle(x):
...
 
+def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
+    """ Reduce a LazyTensor along an axis with function fun using batches.
+
+    When axis=None, reduce the LazyTensor to a scalar as a sum of fun over
+    batches taken along dim.
+
+    Parameters
+    ----------
+
```
agramfort, line 503:
```suggestion
```

ot/utils.py
```diff
@@ -492,6 +492,79 @@ def get_coordinate_circle(x):
...
+        If 0 or 1 reduce the LazyTensor to a vector/matrix as a sum of fun over
+        batches taken along axis.
+    nx : Backend, optional
+        Backend to use for the reduction
+    batch_size : int, optional
+        Size of the batches to use for the reduction (default=100)
+
+    Returns
+    -------
+
```
agramfort, line 520:
```suggestion
```

ot/utils.py
```diff
@@ -492,6 +492,79 @@ def get_coordinate_circle(x):
...
+    res : array-like
+        Result of the reduction
+
+    """
+
+    if nx is None:
+        nx = get_backend(a[0])
+
+    if batch_size is None:
+        batch_size = 100
```
agramfort, line 530: make 100 the default in signature

ot/utils.py
```diff
@@ -492,6 +492,79 @@ def get_coordinate_circle(x):
...
+
+    When axis=None, reduce the LazyTensor to a scalar as a sum of fun over
+    batches taken along dim.
+
+    Parameters
+    ----------
+
+    a : LazyTensor
+        LazyTensor to reduce
+    fun : callable
```
agramfort, line 506: I would call this func rather than fun

ot/utils.py
```diff
@@ -492,6 +492,79 @@ def get_coordinate_circle(x):
     return x_t
 
 
+def reduce_lazytensor(a, fun, axis=None, nx=None, batch_size=None):
+    """ Reduce a LazyTensor along an axis with function fun using batches.
```
agramfort, line 496: the code it's just for 2d arrays/tensors. Do you confirm?

rflamary: updated the doc, t should work for any shape but reuction is imllemented only for global and axis 1 and 2

Issue #530, opened 2023-10-09
gromov_barycenters function always returns the zero matrix


youssef62:
## Describe the bug

`gromov_barycenters` function always returns the zero matrix as cost matrix.
I executed the code below multiple times and it always returns 0.
#### Screenshots

<img width="582" alt="image" src="https://github.com/PythonOT/POT/assets/29061556/f470f09f-d888-4c0a-a772-42973e510e82">

#### Code sample

```python
from ot.gromov import gromov_barycenters

n = 5
g1 = nx.erdos_renyi_graph(n,.8)
g2 = nx.erdos_renyi_graph(n,.8)

c1 = nx.adjacency_matrix(g1).toarray()
c2 = nx.adjacency_matrix(g2).toarray()
print(c1)
print(c2) 
b , log = gromov_barycenters(n,[c1,c2] ,log = True )  
print(b)
print(log)
```

### Expected behavior
A non zero matrix.

### Environment (please complete the following information):
- OS (e.g. MacOS, Windows, Linux): MacOs , also tried on collab
- Python version:Python 3.11.5
- How was POT installed (source, `pip`, `conda`): pip 

Output of the following code snippet:
```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import ot; print("POT", ot.__version__)
```
```
macOS-12.5.1-arm64-arm-64bit
Python 3.11.5 (main, Aug 24 2023, 15:09:32) [Clang 14.0.0 (clang-1400.0.29.202)]
NumPy 1.26.0
SciPy 1.11.3
POT 0.9.1
```

cedricvincentcuaz:
Hello @youssef62 ,

The problem comes from the fact that 'c1 = nx.adjacency_matrix(g1).toarray()' as by default 'int64' as type. So the gromov wasserstein solver called several times in 'gromov_barycenters' will always convert the outputted transport plan to 'int64' i.e a 0 matrix.
You can simply fix your code on your side by doing e.g  'c1 = nx.adjacency_matrix(g1).toarray().astype(np.float64)'.

On our side, we could fix this bug by requesting to provide float inputs at least, no matter the backend.

rflamary:
[...]

PR #560, opened 2023-11-04
[WIP] documentation and warnings in (f)gw cg solvers for integer inputs

cedricvincentcuaz:
## Types of changes

- Add note for integer inputs in the documentation of :`gromov_wasserstein`, `gromov_wasserstein2`, `fused_gromov_wasserstein`, `fused_gromov_wasserstein2` 
- Add warnings for integer inputs in : `gromov_wasserstein`, `fused_gromov_wasserstein`

- Address issue #530 

- Add tests with integers in `test_gromov`: `test_gromov_integer_warnings`, `test_fgw_integer_warnings`

## PR checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -19,7 +19,7 @@
 #### Closed issues
 - Fix line search evaluating cost outside of the interpolation range (Issue #502, PR #504)
 - Lazily instantiate backends to avoid unnecessary GPU memory pre-allocations on package import (Issue #516, PR #520)
-
+- Handle documentation and warnings when integers are provided to (f)gw solvers based on cg (Issue #530, PR #559)
 
 ## 0.9.1
 *August 2023*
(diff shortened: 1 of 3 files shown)


cedricvincentcuaz:
Hello @rflamary,

For GW (resp. FGW) solvers I set the type of the first variable $C_1$ (resp. $M$) as default type for the outputted transport plan. Do you think that it is enough ? Would it be better to make $T$ inherit the most precise type over all inputs ?
Some use cases: If $p$ and $q$ are not provided there will be instantiated based on this type so the transport plan will inherit it as $G_0 = pq^\top$. If $p$ or $q$ are provided as integers, $G_0$ would have int type so most likely would be a matrix of 0. But providing such marginals really does not make sense. Same argument if $G_0$ is provided by the user with an int type.

rflamary:
Most precise type of the input is not possible to find easily because of the backend. Its OK to keep the type of M (if aailable) or C with a warning as long as it is clearly stated in thedocumentation.

Issue #565, opened 2023-11-07
CUDA out of memory when using ot.sinkhorn2 as a loss function


Perian-Yan:
Hi, I'm trying to implement emd/sinkhorn distance as the loss function for 2D matrices. 
However, `ot.sinkhorn2` causes CUDA out of memory error when it's being computed:

![image](https://github.com/PythonOT/POT/assets/100858995/14069057-ae8f-4f9d-94d7-6b9816abd9e3)

The `ot.emd2` can also give this error when I use a larger data set. 

By viewing `nvidia-smi`, when I try to train the same dataset, `ot.emd2` uses up about 7G/12G memory (which is fine), while `ot.sinkhorn2` uses 11G/12G memory and causes the error above.

Thank you!

rflamary:
Hello, 

This  probably comes form the fact that torch needs to keep the intermediate values in sinkhorn in memory to allow for a backward(). I have been planning to add a detach_iterations parameter for a hile that will allow to run the whole algorithm and plug the gradient assuming convrgence (implicit differettiation) for a while. I will get to that when I have more time. 

Could you please check if sinkhorn2 explodes in memory when you give it arrays with keep_gradient=False?

rflamary:
Hello, I have added an option to `ot.solve`that forces the use of implicit gradients and limits the memory use:

you can compute the sinkhorn loss with
```python
loss = ot.solve(M, a, b, grad='implicit').value
```
All iterations are detached and the gradient is set at the end with no memory overhead but then it is differentiable only wrt value (not value_linear or the OT plan). Could you tell me if it solves your problem, it is merges in master brach ?

Guddubhaiya07:
**This probably comes form the fact that torch _needs to keep the intermediate _values in sinkhorn in memory to allow for a backward(). I have been planning to add a detach_iterations parameter for a hile that will allow to run the whole algorithm and plug the gradient assuming convrgence - [ ] # (implicit differettiation) for a while. I will_` get to that when I have more time.****

cedricvincentcuaz:
This new functionality is now available in `ot.solve` and `ot.solve_sample` thanks to PR #605 .

PR #605, opened 2024-02-19
[MRG] Add implicit Sinkhorn gradients

rflamary:
## Types of changes

This PR aims at 
- [x] implementing the detach function in the backend to allow speedup on CPU/GPU in some solvers (which was already done in a previous PR but with limited doc).
- [x] Implement variants of Sinkhorn where computations are detached and gradients at convergence is returned instead

This PR should solve #565 and greatly limit memory for sinkhorn when computing gradienst wrt the value.

In order to use implicit diffeerntiation one needs to set the `grad` parameter in `ot.solve`and `ot.solve_sample`as such
```python
sol = ot.solve(M, a, b, reg=10, grad='implicit')
sol.value.backward()
# beware with  grad='implicit', sol.value_linear and sol.plan are not differentiable (not implemented yet).
```
On a simple example with pytorch arrays with required gradients, I has a 1000x gain in memory for solving the problem when a  large number of sinkhorn operations are needed.

## PR checklist

### DIFF
--- a/ot/backend.py
+++ b/ot/backend.py
@@ -281,12 +281,12 @@ class Backend():
         """Define the gradients for the value val wrt the inputs """
         raise NotImplementedError()
 
-    def detach(self, a):
+    def detach(self, *arrays):
         """Detach the tensor from the computation graph"""
         if len(arrays) == 1:
-            return self._detach(arrays[0], type_as=type_as)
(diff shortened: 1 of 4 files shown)


cedricvincentcuaz (approved):
Hello Rémi ! Ready to merge ;)

PR #607, opened 2024-02-29
[MRG] Faster gromov-wasserstein linesearch for  symmetric matrices

KrzakalaPaul:
## Types of changes

Add a new argument "symmetric" to ot.gromov._gw.solve_gromov_linesearch. The default value symmetric=False results in the previous behaviour. When C1 and C2 are symmetric however, one can set symmetric=True which results in faster computations as the formula can be simplified in that case. 

ot.gromov._gw.fused_gromov_wasserstein and ot.gromov._gw.gromov_wasserstein have been changed accordingly to automatically pass symmetric=False or symmetric=True to solve_gromov_linesearch.

A new file symmetric_linesearch.py has been added to examples.gromov to check the usefullness of setting symmetric=True. 

The formula for solving the gromov linesearch can be simplified in the (very common case) where the matrices C1 and C2 are symmetric, resulting in a small (but free) improvement in computationnal cost. 

## PR checklist


cedricvincentcuaz (changes requested):
Good catch @KrzakalaPaul , thank you :)
Could you edit the `releases.md` with your PR ? then the PR will be ready for merging.

Best, Cédric.

KrzakalaPaul:
Hi :)
I tried to edit releases.md, tell me if it's alright.
Best, Paul

PR #610, opened 2024-03-01
fix the sign of gradient for kl gromov

KrzakalaPaul:
## Types of changes
Correct a sign error of the nx.set_gradient for gromov (and fused gromov) when loss_fun = 'kl_loss'.
The correct formula is:

gC2 = - nx.dot(T.T, nx.dot(C1, T)) / (C2 + 1e-15) + nx.outer(q, q)

instead of 

gC2 =  nx.dot(T.T, nx.dot(C1, T)) / (C2 + 1e-15) + nx.outer(q, q)

You can run this code:
```
from ot import gromov_wasserstein2,unif
import torch 

C1 = torch.rand((2,2), requires_grad = False)
C2 = 1 - C1
C2.requires_grad = True

eta = 1e-1

for step in range(100):
    
    loss = gromov_wasserstein2(C1=C1,C2=C2,p=unif(2,type_as=C2),q=unif(2,type_as=C2),loss_fun='square_loss')
    grad = torch.autograd.grad(loss, C2)[0]
    C2 = C2 - eta*grad
    C2 = torch.clip(C2,0,1)
    
    print(loss)
````

You will see that the gradient descent diverges. It converges when we fix the sign error.

## PR checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,7 +10,7 @@
 - Fix gpu compatibility of sr(F)GW solvers when `G0 is not None`(PR #596)
 - Fix doc and example for lowrank sinkhorn (PR #601)
 - Fix issue with empty weights for `ot.emd2` (PR #606, Issue #534)
-- Fix a sign error regarding the gradient of `ot.gromov._gw.fused_gromov_wasserstein2` and `ot.gromov._gw.gromov_wasserstein2` for the kl loss
+- Fix a sign error regarding the gradient of `ot.gromov._gw.fused_gromov_wasserstein2` and `ot.gromov._gw.gromov_wasserstein2` for the kl loss (PR #610)
 
 ## 0.9.2
 *December 2023*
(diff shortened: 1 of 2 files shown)


cedricvincentcuaz:
Hello @KrzakalaPaul, indeed good catch. 

In your example you do not have `loss_fun = kl_loss`. So doing a quick check to be sure, in this case the GW loss reads as
$E(\mathbf{A}, \mathbf{B}, \mathbf{T}) = \sum_{ijkl} KL(A_{ij}, B_{kl})T_{ik}T_{jl}$
with $KL(A_{ij}, B_{kl}) = A_{ij} \log(A_{ij}) - A_{ij}\log(B_{kl}) - A_{ij} + B_{kl}$.

So $\frac{\partial E}{\partial B_{pq}} = \sum_{ij} (- \frac{A_{ij}}{B_{kl}} + 1) T_{ip}T_{jq}$ and we indeed forgot a `-` in the current POT implementation.

PR #623, opened 2024-05-17
fix: check for NaNs in emd loss matrix

### LINKED ISSUE #469: nan in metric cost matrix causes 'segmentation fault core dumped'
## Bug

Hello!
When there is a nan in the metric cost matrix (gromov_wasserstein or fused_gromov_wasserstein) I get a 'segmentation fault core dumped' error. 

## Code sample

import numpy as np
from ot import gromov_wasserstein2

C1 = np.zeros((2,2))
C2 = np.ones((2,2))
p = np.array([0.5,0.5])
q = np.array([0.5,0.5])

gromov_wasserstein2(C1, C2, p, q)

C1[0,0]=np.nan

gromov_wasserstein2(C1, C2, p, q)

## Expected behavior

I would expect a more clear warning/error or even that gromov_wasserstein2 returns a nan.

bobluppes:
## Types of changes
This PR introduces an additional check for `NaN`s in the loss matrix of the emd computation. If `NaN`s are detected we raise an error in order to protect against segfaults in the C++ backend.

The motivation of this PR is to fail more gracefully in cases of `NaN` costs. 
Closes #469 

Added new tests.

## PR checklist
- [ ] I have read the [**CONTRIBUTING**](CONTRIBUTING.md) document.

> ⚠️ 
Some notes on the checklist above:
- I did not find a `CONTRIBUTING.md`
- While I ran all related tests, I did not run the entire suite (due to some modules missing on my system). I assume the entire suite is run in the CI?
- Please let me know if this is something you would like a release issue for. If so please let me know.

### DIFF
--- a/ot/lp/__init__.py
+++ b/ot/lp/__init__.py
@@ -237,6 +237,8 @@ def emd(a, b, M, numItermax=100000, log=False, center_dual=True, numThreads=1, c
 
     .. note:: An error will be raised if the vectors :math:`\mathbf{a}` and :math:`\mathbf{b}` do not sum to the same value.
 
+    .. note:: An error will be raided if the loss matrix :math:`\mathbf{M}` contains NaNs.
+
     Uses the algorithm proposed in :ref:`[1] <references-emd>`.
 
     Parameters
@@ -301,6 +303,9 @@ def emd(a, b, M, numItermax=100000, log=False, center_dual=True, numThreads=1, c
     ot.bregman.sinkhorn : Entropic regularized OT
     ot.optim.cg : General regularized OT
     """
+
+    if np.isnan(M).any():
+        raise ValueError('The loss matrix should not contain NaN values.')
 
     a, b, M = list_to_array(a, b, M)
     nx = get_backend(M, a, b)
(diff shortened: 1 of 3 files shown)


ot/lp/__init__.py
```diff
@@ -302,6 +304,9 @@ def emd(a, b, M, numItermax=100000, log=False, center_dual=True, numThreads=1, c
     ot.optim.cg : General regularized OT
     """
 
+    if np.isnan(M).any():
+        raise ValueError('The loss matrix should not contain NaN values.')
+
```
bobluppes, line 309: Failing early here ensures that we do not segfault in the accelerated `emd_c` call. 

I did not look too deep into the `emd_c` implementation, but my assumption is that this check is somewhat pessimistic. Maybe it is possible to formulate problems for which we do not need to access a subset of values in the loss matrix (possibly due to the graph being disconnected). In that case we could support NaN values in some cases. @rflamary what is your opinion on this?

rflamary: if the graph is disconnected then the parts that are not used should have an infinite value (which is ha,ndled by the C++ solver). i'm OK with not handling naNs.

ot/lp/__init__.py
```diff
@@ -302,6 +304,9 @@ def emd(a, b, M, numItermax=100000, log=False, center_dual=True, numThreads=1, c
     ot.optim.cg : General regularized OT
     """
 
+    if np.isnan(M).any():
```
rflamary, line 307: A problem here is that you are using numpy on arrays that might not be numpy (see backend function below). You should do the test later in the function on the OT loss marix that hhas been converted to numpy to avoid backend errors.

PR #657, opened 2024-07-16
Use raw strings with TeX

gspr:
Some instances of docstrings using TeX were incorrectly specified as ordinary strings, leading to the TeX macros being interpreted as escape sequences.


cedricvincentcuaz:
Good catch, thank you for the PR @gspr

PR #676, opened 2024-09-13
[MRG] Translation Invariant Sinkhorn for Unbalanced OT

clbonet:
## Types of changes

This PR aims to add the Translation Invariant Sinkhorn algorithm proposed in [Faster Unbalanced Optimal Transport: Translation invariant Sinkhorn and 1-D Frank-Wolfe
](https://proceedings.mlr.press/v151/sejourne22a.html) to solve the Unbalanced OT problem. 

- Added the function `sinkhorn_unbalanced_translation_invariant`  in `unbalanced/_sinkhorn.py`
- Added an example in `plot_conv_sinkhorn_ti.py` to demonstrate the convergence benefits
- Added the method in `unbalanced/test_sinkhorn.py`

This version of Sinkhorn converges faster than the classical Sinkhorn algorithm for UOT.

The function can be called by using the method "sinkhorn_translation_invariant" in `sinkhorn_unbalanced`, so I added this method in the itertools of tests of `unbalanced/test_sinkhorn.py`.

## PR checklist

### DIFF
--- a/ot/unbalanced/_sinkhorn.py
+++ b/ot/unbalanced/_sinkhorn.py
@@ -362,6 +362,9 @@ def sinkhorn_unbalanced2(a, b, M, reg, reg_m, method='sinkhorn',
             return cost
 
     else:
+        if reg_type == "kl":
+            warnings.warn('Reg_type not implemented yet. Use entropy.')
+
         if method.lower() == 'sinkhorn':
             return sinkhorn_knopp_unbalanced(a, b, M, reg, reg_m, reg_type, c,
                                              warmstart, numItermax=numItermax,
--- a/ot/unbalanced/_sinkhorn.py
+++ b/ot/unbalanced/_sinkhorn.py
@@ -372,10 +372,10 @@ def sinkhorn_unbalanced2(a, b, M, reg, reg_m, method='sinkhorn',
                                                   log=log, **kwargs)
 
         elif method.lower() == 'sinkhorn_translation_invariant':
-            return sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, reg_type, c,
-                                                  warmstart, numItermax=numItermax,
-                                                  stopThr=stopThr, verbose=verbose,
-                                                  log=log, **kwargs)
+            return sinkhorn_unbalanced_translation_invariant(a, b, M, reg, reg_m, reg_type, c,
+                                                             warmstart, numItermax=numItermax,
+                                                             stopThr=stopThr, verbose=verbose,
+                                                             log=log, **kwargs)
 
         elif method.lower() in ['sinkhorn_reg_scaling']:
             warnings.warn('Method not implemented yet. Using classic Sinkhorn-Knopp')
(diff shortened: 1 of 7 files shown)


ot/unbalanced/_sinkhorn.py
```diff
@@ -865,6 +895,252 @@ def sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, reg_type="kl", c=None,
...
+    k2 = reg * reg_m2 / ((reg + reg_m2) * (reg_m1 + reg_m2)) if reg_m2 != float("inf") else 0
+
+    k_rho1 = k1 * reg_m1 / reg if reg_m1 != float("inf") else 0
+    k_rho2 = k2 * reg_m2 / reg if reg_m2 != float("inf") else 0
+
+    if reg_m1 == float("inf") and reg_m2 == float("inf"):
+        xi1, xi2 = 0, 0
+        fi_12 = 1
+    elif reg_m1 == float("inf"):
+        xi1 = 0
```
rflamary, line 1059: Why are these line sot tested? imaybe the test shoudl run with infinite values too

ot/unbalanced/_sinkhorn.py
```diff
@@ -865,6 +895,252 @@ def sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, reg_type="kl", c=None,
...
+        xi2 = reg / reg_m2
+        fi_12 = reg_m2
+    elif reg_m2 == float("inf"):
+        xi1 = reg / reg_m1
+        xi2 = 0
+        fi_12 = reg_m1
```
rflamary, line 1065: same here

ot/unbalanced/_sinkhorn.py
```diff
@@ -865,6 +895,252 @@ def sinkhorn_stabilized_unbalanced(a, b, M, reg, reg_m, reg_type="kl", c=None,
...
+    if log:
+        dict_log['logu'] = nx.log(u + 1e-300)
+        dict_log['logv'] = nx.log(v + 1e-300)
+
+    if n_hists:  # return only loss
+        res = nx.einsum('ik,ij,jk,ij->k', u, K, v, M)
+        if log:
+            return res, dict_log
+        else:
+            return res
```
rflamary, line 1123: this line was not tested

rflamary (commented):
This is very ice thanks @clbonet ,

I have two comments about missing tests since you code handle infinite values it should be tested.

ot/unbalanced/_sinkhorn.py
clbonet: I think the test only handled ("inf", "inf"). I added a test to handle the semi-relaxed case.

ot/unbalanced/_sinkhorn.py
clbonet: I called the wrong function in `sinkhorn_unbalanced2`. It should be ok now.

clbonet: This is now tested in the function `test_unbalanced_multiple_inputs`, but it seems there is a problem with `sinkhorn_unbalanced2` when there is only one histogram b. It does not pass the test for any of the methods.

clbonet: It is ok now. The problem was that the reg_type "kl" is not implemented yet for multiple inputs.

rflamary:
Hello @clbonet ,

Sorry but we did another merging and now there is a conflict with your PR (only minor things such as reference number in redame and release file). Could you please have a quick look (and also update reference numbers in the doc of functions) so that we can finally merge

PR #683, opened 2024-10-31
[WIP] Sparse implementation of EMD

tvayer:
The goal is to return a sparse transport in `ot.emd` plan when the cost matrix matrix is itself sparse which corresponds to infinite cost between points, or when no value is given i.e. we know that no match is allowed.

### Wishlist
- Accelerate the EMD  when the cost is sparse
- Stay in O(nnz(C)) in memory

### What we have
- Working cython wrapper for coo cost matrix, it returns the same result at numerical precision

### What we want 
- Change the cpp of NetworkSimplex to remove the O(n^2) memory when associating the cost to each edge of the network flow


rflamary:
This has been done by @nathanneike thanks @tvayer for the work but 'im closing the PR

PR #692, opened 2024-11-18
[MRG] no longer impose ECOS solver for SSNB, use new CVXPY default (Clarabel)

### LINKED ISSUE #688: CVXPY >= 1.6.0 removes its dependency to ECOS
## Describe the bug

Current POT tests fail because of a change in dependencies of CVXPY (see [updates](https://www.cvxpy.org/updates/) here)

#### Screenshots

Screenshot 2024-11-16 at 12 47 49 PM

eloitanguy:
## Types of changes
CVXPY changed its default solver (from ECOS to Clarabel), changing the mapping solvers to use the default CVXPY solver instead of asking for an old solver that is no longer installed by default

Fix #688

Example + pytest

## PR checklist

- [x]  I have read the [**CONTRIBUTING**](https://pythonot.github.io/contributing.html) document.

### DIFF
--- a/ot/mapping.py
+++ b/ot/mapping.py
@@ -173,7 +173,7 @@ def nearest_brenier_potential_fit(
                         - c3 * (G[j] - G[i]).T @ (X[j] - X[i])
                     ]
         problem = cvx.Problem(objective, constraints)
-        problem.solve(solver=cvx.ECOS)
+        problem.solve()
         phi_val, G_val = phi.value, G.value
         it_log_dict = {
             "solve_time": problem.solver_stats.solve_time,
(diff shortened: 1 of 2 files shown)


ot/mapping.py
```diff
@@ -173,7 +173,7 @@ def nearest_brenier_potential_fit(
                         - c3 * (G[j] - G[i]).T @ (X[j] - X[i])
                     ]
         problem = cvx.Problem(objective, constraints)
-        problem.solve(solver=cvx.ECOS)
+        problem.solve()
```
rflamary, line 176: could be good to allow the solver to be passed in parameter? My experience anis that it can have a large impact on performance

eloitanguy: sounds good, and we'll leave the default as the cvxpy default

Issue #723, opened 2025-03-12
entropic_partial_wasserstein not stable


wzm2256:
## Describe the bug
The entropic_partial_wasserstein function produces nan when eps is small.

### To Reproduce
```
import ot
import torch
import numpy as np

def compute_OT(M, alpha, beta, epsilon):
    s1, s2 = M.shape[0], M.shape[1]
    assert s1 == s2
    unif_vec = ot.unif(s1)
    a, b = unif_vec/beta, unif_vec
    pi_1_np = ot.partial.entropic_partial_wasserstein(a, b, M, m=alpha, reg=epsilon)
    print(f"Original: sum(pi) = {pi_1_np.sum():.4f}, alpha = {alpha:.4f}")


beta = 0.35
alpha = 0.01


M_1 = torch.load('M_1.pt')
print(f"M_1 norm = {np.linalg.norm(M_1):.2f}\n")

epsilon = 10.
compute_OT(M_1, alpha, beta, epsilon)

epsilon = 0.1
compute_OT(M_1, alpha, beta, epsilon)

```

#### Output
```
Original: sum(pi) = 0.0100, alpha = 0.0100
G:\Mycode\POT\ot\partial.py:698: RuntimeWarning: divide by zero encountered in divide 
  np.multiply(K, m / np.sum(K), out=K)
G:\Mycode\POT\ot\partial.py:698: RuntimeWarning: invalid value encountered in multiply
  np.multiply(K, m / np.sum(K), out=K)
Warning: numerical errors at iteration 0
Original: sum(pi) = nan, alpha = 0.0100
```
When eps=0.1, the output is Nan.

### Environment (please complete the following information):
- OS (e.g. MacOS, Windows, Linux): Windows
- Python version: 3.10
- How was POT installed (source, `pip`, `conda`): pip
- Build command you used (if compiling from source):
- Only for GPU related bugs:
    - CUDA version:
    - GPU models and configuration:
    - Any other relevant information:

Output of the following code snippet:
```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import ot; print("POT", ot.__version__)
```

wzm2256:
I will create a pull request soon.

rflamary:
Closing since fixed in #811

PR #724, opened 2025-03-13
[WIP] Add a stablized function entropic_partial_wasserstein_logscale

wzm2256:
1. add a new function called entropic\_partial\_wasserstein\_logscale to ot.partial. This function solves the same problem as entropic\_partial\_wasserstein but is computed in logscale, so it is more robust.

2. Test exampless are provided in compare_logscale_POT.py. Test data is in data\entropic_partial_OT_cost.txt

## Types of changes

I implement a new function `entropic_partial_wasserstein_logscale` that solves exactly the same problem as the one in `entropic_partial_wasserstein` in log scale. The new function is a line-to-line translation of the old one, and the input/output format is exactly the same.

I do not remove the old function because the new function can be slower due to the use of the logsumexp trick. So when there is no Nan error, the old function is favored.

https://github.com/PythonOT/POT/issues/723

I test the new function `entropic_partial_wasserstein_logscale` against the old one `entropic_partial_wasserstein` in the example file 'compare_logscale_POT.py` for both numpy and pytorch.
1. When eps is large, the old and new function produces exactly the same result 
2. When eps is small, the old function produce an error, but the new function works well. 

## PR checklist

I could not build the document in my laptop due to some errors:
```
Theme error:
no theme named 'sphinx_rtd_theme' found (missing theme.toml?)
```
 so I am not completely sure whether the document is fine, although I only added a few sentences to the docs.

Also, I do not know how to use pytest to test my code. If this is necessary, I may need some help here.


rflamary:
Hello @wzm2256 and thanks for the PR we are indeed missing a stabilized partial entropic ot solver.

When we add a stabilized solver, we usually add the function and make the original function as a wrapper when we add the log implementation (see how it is done for function ot.sinkhorn ) so that with an additional parameter `method='sikhorn_log'` one can select the stabilized solver. Could you please do that and add the tests ?

Finally your example is nice we usually try to add a visualization and we really try not to add raw data to the git. Could you design a simulated example instead of a new dataset?

Thanks again for your work, we will do a proper code review as soon as possible but wold prefer to do that after the comments above are taken into account.

wzm2256:
Sure! I will work on that.

rflamary:
Hello @wzm2256 , quick reminder that we are waiting for a few changes here.

hinanohart:
@wzm2256 @rflamary @cedricvincentcuaz — hello! I'm a downstream user of `entropic_partial_wasserstein` and hit exactly the NaN regime described in #723. I've prepared the missing pieces (pytest cases, Sphinx-Gallery example, docs blurb, `RELEASES.md`) as a rescue follow-up at #811.

Why a separate PR instead of pushing here: `ot/partial.py` was split into the `ot/partial/` package after this PR opened, so the diff against current `master` is now mostly unrelated movement. Re-applying just @wzm2256's function on the new layout is cleaner than rebasing through that split.

I'm happy to do whichever you prefer:

1. **Keep #811 as the canonical rescue PR** (default) — wzm2256 retained as algorithm author in the description.
2. **Close #811 and push these commits to this branch** — if @wzm2256 grants write access, or if @rflamary / @cedricvincentcuaz prefers to keep #724 as the canonical PR.

Either way, the goal is to get the function in so issue #723 closes. Let me know which path you want.

rflamary:
Closing this because #811 was merged

PR #736, opened 2025-05-10
[MRG] Linear Circular OT

clbonet:
## Types of changes

This PR aims to add the Linear Circular OT distance (with uniform measure on the circle as reference measure) introduced in [LCOT: Linear Circular Optimal Transport
](https://openreview.net/forum?id=49z97Y9lMq) and its sliced counterpart introduced in [Linear Spherical Sliced Optimal Transport: A Fast Metric for Comparing Spherical Data
](https://openreview.net/forum?id=fgUFZAxywx).

### Changes:
- Added the `ot.solver_1d.linear_circular_embedding` and `ot.solver_1d.linear_circular_ot` functions to compute the Linear Circular OT distance.
- Added the `ot.sliced.linear_sliced_wasserstein_sphere` function to compute its sliced counterpart.
- Refactored the code for the spherical sliced distances, adding the functions `ot.sliced.get_projections_sphere` and `ot.sliced.projection_sphere_to_circle`.
- Added tests for these functions in `test/test_1d_solver` and `test/test_sliced`.
- Added the Linear Circular OT distance in the example `plot_compute_wasserstein_circle.py`
- Modified the default behavior of `ot.wasserstein_circle` to always use `ot.solver_1d.binary_search_circle` for any p, which seems to give better results even for p=1 (as discussed in Issue #738).
- Improved the visualization of the sphere in the example `plot_ssw_unif_torch.py`.

The Linear Circular OT distance with uniform measure on the circle is very fast to compute, as the OT maps can be computed in closed-forms. It is thus faster to compute than the true wasserstein distance on the circle.

I added tests of these functions in `test/test_1d_solver` and `test/test_sliced`.

## PR checklist


examples/plot_compute_wasserstein_circle.py
```diff
@@ -159,15 +172,23 @@ def pdf_von_Mises(theta, mu, kappa):
...
+L_lcot = np.zeros((n_try, 100))
 for i in range(n_try):
     L_w2[i] = ot.semidiscrete_wasserstein2_unif_circle(xts[i].T)
+    L_lcot[i] = ot.linear_circular_ot(xts[i].T)
 
 m_w2 = np.mean(L_w2, axis=0)
 std_w2 = np.std(L_w2, axis=0)
 
+m_lcot = np.mean(L_lcot, axis=0)
+std_lcot = np.mean(L_lcot, axis=0)
```
rflamary:
```suggestion
std_lcot = np.std(L_lcot, axis=0)
```

ot/lp/solver_1d.py
```diff
@@ -791,10 +791,12 @@ def binary_search_circle(
                 -1, 1
             )
 
-            mask_end = mask * (nx.abs(dCptm - dCmtp) > 0.001)
-            tc[mask_end > 0] = (
-                (Ctp - Ctm + tm * dCptm - tp * dCmtp) / (dCptm - dCmtp)
-            )[mask_end > 0]
+            with warnings.catch_warnings():
```
rflamary, line 797: this is a bit worrying. why are you catching tjose worning? when do they happen. It is probaly OK but we need at last a comment there to explian

ot/lp/solver_1d.py
```diff
@@ -996,17 +998,19 @@ def wasserstein_circle(
...
     .. [44] Hundrieser, Shayan, Marcel Klatt, and Axel Munk. "The statistics of circular optimal transport." Directional Statistics for Innovative Applications: A Bicentennial Tribute to Florence Nightingale. Singapore: Springer Nature Singapore, 2022. 57-82.
     .. [45] Delon, Julie, Julien Salomon, and Andrei Sobolevski. "Fast transport optimization for Monge costs on the circle." SIAM Journal on Applied Mathematics 70.7 (2010): 2239-2258.
     """
     assert p >= 1, "The OT loss is only valid for p>=1, {p} was given".format(p=p)
 
-    if p == 1:
-        return wasserstein1_circle(
-            u_values, v_values, u_weights, v_weights, require_sort
-        )
+    # if p == 1:
```
rflamary: remove the code if not usefull anymore

ot/lp/solver_1d.py
```diff
@@ -1095,3 +1099,150 @@ def semidiscrete_wasserstein2_unif_circle(u_values, u_weights=None):
...
+    v_values : ndarray, shape (n, ...), optional
+        samples in the target domain (coordinates on [0,1[), if None, compute distance against uniform distribution
+    u_weights : ndarray, shape (n, ...), optional
+        samples weights in the source domain
+    v_weights : ndarray, shape (n, ...), optional
+        samples weights in the target domain
+
+    Returns
+    -------
+    loss: float
```
rflamary: is this true or is this of shape (...) instead?

rflamary (commented):
This is a very strong PR, thanks again @clbonet !

Just a few comments and I think we can merge. Also you new function get_projections_sphere is very nice, thansk for teh factoring

ot/lp/solver_1d.py
clbonet: It is of shape (...)! I will correct it, thank you!

ot/lp/solver_1d.py
clbonet: These warnings are raised when  some elements of `dCptm - dCmtp` are 0. But in this case, the array `tc` is not updated because `mask_end` is False.

So I thought that it should be better to not raise them. There are maybe other solutions to avoid these warnings (I tried to put the masks insides the operations, but it broke the shapes, so I tried with the catch).

I will add a comment.

PR #741, opened 2025-07-01
[MRG] Partial optimal transport 1d solver

rtavenar:
...from Chapel & Tavenard 2025

## Types of changes

It introduces an efficient partial optimal transport solver for 1d problems

Basic tests on dimensions of outputs, and the main test consists in checking that this solver has similar outputs to the already existing partial wasserstein solver applied to 1d data.

## PR checklist


rtavenar:
OK I have something I am rather happy with. My last step would be to move from list/dict/set implementation to only numpy arrays and benchmark efficiency.

@rflamary maybe you can have a look at the PR before this last step, as that should not impact too much the overall integration of the method in the package (docs, example and API should stay the same).

Also, I'm not sure when I will have time for this last step (should either be tomorrow or in 1-2months).

setup.py
```diff
@@ -62,17 +62,26 @@
...
+                language="c++",
+                include_dirs=[numpy.get_include(), os.path.join(ROOT, "ot/lp")],
+                extra_compile_args=compile_args,
+                extra_link_args=link_args,
+            ),
+            Extension(
+                name="ot.partial.partial_cython",
+                sources=["ot/partial/partial_cython.pyx"],
+                include_dirs=[numpy.get_include(), os.path.join(ROOT, "ot/partial")],
+                extra_compile_args=["-O3"],
```
rflamary: O3 already given line 43, why only - here onthis might be  aporblem on darwin

rflamary (commented):
This is very nice thank @rtavenar 

I have one comment about the compile args?

setup.py
rtavenar: Oh right, I didn't notice the `compile_args` var.

PR #793, opened 2026-02-09
[MRG] Warmstart with dual potentials for exact OT solver

nathanneike:
…omputed potentials from sinkhorn or even related simplex

## Types of changes

Add Warmstart intialisation to the simplex solver allows us to give good potentials to our solvers and get why faster to solution with less iterations. 

This change allows us to warmstart with potentials from other solvers. Instead of using normal initial pivots to create a Tree. We use the given potential to create a first feasible solution in our Simplex. If given potentials are good potential for big speedup. Still need to compute all Reduce Costs at least once which means that the running time for the warmstart can be big for big problems so should be done with good potentials and not just any potentials. 

Added a test for verifying correctness. 
Speedup for exact potentials from an exact solve is around 3x and using good sinkhorn potentials 1.4x.

## PR checklist

### DIFF
--- a/ot/lp/_network_simplex.py
+++ b/ot/lp/_network_simplex.py
@@ -238,7 +238,7 @@ def emd(
     check_marginals: bool, optional (default=True)
         If True, checks that the marginals mass are equal. If False, skips the
         check.
-    warmstart_dual: tuple of two arrays (alpha, beta), optional (default=None)
+    potentials_init: tuple of two arrays (alpha, beta), optional (default=None)
         Warmstart dual potentials to accelerate convergence. Should be a tuple
         (alpha, beta) where alpha is shape (ns,) and beta is shape (nt,).
         These potentials are used to guide initial pivots in the network simplex.


ot/lp/_network_simplex.py
```diff
@@ -513,6 +529,11 @@ def emd2(
     check_marginals: bool, optional (default=True)
         If True, checks that the marginals mass are equal. If False, skips the
         check.
+    warmstart_dual: tuple of two arrays (alpha, beta), optional (default=None)
```
rflamary: added to doc but not in teh function params and the call no?

test/test_ot.py
```diff
@@ -1126,6 +1126,37 @@ def test_emd2_sparse_vs_dense_gradients():
     assert np.abs(grad_values.sum() - 1.0) < 1e-7
 
 
+def test_emd_warmstart():
```
rflamary: test_emd_dual_warmstart

test/test_ot.py
```diff
@@ -1126,6 +1126,37 @@ def test_emd2_sparse_vs_dense_gradients():
...
+    # run solver with limited iterations (stop early)
+    G_partial, log_partial = ot.emd(a, b, M, numItermax=100, log=True)
+    u_partial = log_partial["u"]
+    v_partial = log_partial["v"]
+
+    # resume from the partial duals with full iterations
+    G_warm, log_warm = ot.emd(a, b, M, log=True, warmstart_dual=(u_partial, v_partial))
+
+    # cold-start reference
+    G_cold, log_cold = ot.emd(a, b, M, log=True)
```
rflamary, line 1149: also ad emd2 (which was missing)

ot/lp/_network_simplex.py
nathanneike: dont get this one they are there, I use them later to set alpha and beta init

EduardoGRocha:
> Hello @nathanneike ,
> 
> It would be nice that you also provide a few computation times in the description of the PR so that we have an idea of the speedup when using warm start.
> 
> Note that potentials_init was used for warm start in ot.solve and you should use this same also in emd.
> 
> I would also like that you update `ot.solve` and `ot.solve_sample` to pass the warmstart to emd.
> 
> few other comments below

@nathanneike did you benchmark the improvements here? I suspect that finding an initial "good" solution is relatively quick.

**Edit:** For context of why I've been asking so many questions. I've been reworking the code to optimize it. But that would be simpler if we "unentangle" the code a bit

nathanneike:
> > Hello @nathanneike ,
> > It would be nice that you also provide a few computation times in the description of the PR so that we have an idea of the speedup when using warm start.
> > Note that potentials_init was used for warm start in ot.solve and you should use this same also in emd.
> > I would also like that you update `ot.solve` and `ot.solve_sample` to pass the warmstart to emd.
> > few other comments below
> 
> @nathanneike did you benchmark the improvements here? I suspect that finding an initial "good" solution is relatively quick.
> 
> **Edit:** For context of why I've been asking so many questions. I've been reworking the code to optimize it. But that would be simpler if we "unentangle" the code a bit

Yes I benchmarked the improvements (could have been more extensive but I spent some time on it).

PR #796, opened 2026-02-17
[MRG] Optimize simplex setup

nathanneike:
## Types of changes

Optimize the setup phase for the dense solver. Passes cost by reference and makes the solver results public for easier access.

- Read Cost directly using pointer passed from Python for this we have to disable arc mixing and have full support (no 0 demand or supply, python wrapper enforces this)
- Extraction of flow results is now only done on none 0 flows saving a lot of time since the remaining matrix is already 0 initialized

Dense setup phase takes a long time this solves it to a big extent same for result extraction

Timing difference : 
After Optimization 
| Problem Size | Before Fix | After Fix
|--------:|--------------:|---------------:|
|     100 |     0.000672 | 0.000761 |
|     200 |     0.002208 | 0.002287 |
|     500 |     0.017246 | 0.013768 |
|    1000 |     0.073153 | 0.057219 |
|    2000 |     0.314500 | 0.252523 |
|    5000 |     1.962666 | 1.423807 |

## PR checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -25,6 +25,7 @@ This new release adds support for sparse cost matrices and a new lazy EMD solver
 - Clean documentation (PR #787)
 - Fix code coverage (PR #791)
 - Fix test of the version of jax in `ot.backend` (PR #794)
+- Reverting the openmp fix on macOS (PR #789) for macOS (PR #797)
 
 
 ## 0.9.6.post1
--- a/ot/helpers/openmp_helpers.py
+++ b/ot/helpers/openmp_helpers.py
@@ -34,16 +34,13 @@ def get_openmp_flag(compiler):
         omp_flag = ["-fopenmp"]
         if sys.platform.startswith("darwin"):
             omp_flag = ["-Xpreprocessor", "-fopenmp"]
-            omp_flag += ["-I/usr/local/opt/libomp/include"]
             #  Assuming `brew install libomp` on recent macOS
             if os.path.isfile("/opt/homebrew/opt/libomp/lib/libomp.dylib"):
                 omp_flag += ["-I/opt/homebrew/opt/libomp/include"]
-                omp_flag += ["/opt/homebrew/opt/libomp/lib/libomp.dylib"]
             else:
                 # Assuming `brew install libomp` on old macOS
                 if os.path.isfile("/usr/local/opt/libomp/lib/libomp.dylib"):
                     omp_flag += ["-I/usr/local/opt/libomp/include"]
-                    omp_flag += ["/usr/local/opt/libomp/lib/libomp.dylib"]
 
     return omp_flag
 


ot/lp/_network_simplex.py
```diff
@@ -379,19 +379,51 @@ def emd(
             a, b, edge_sources, edge_targets, edge_costs, numItermax
         )
     else:
+        row_mask = asel
```
rflamary, line 425: do not use a new variable for this

ot/lp/_network_simplex.py
```diff
@@ -379,19 +379,51 @@ def emd(
...
+        col_mask = bsel
+        need_filter = np.any(~row_mask) or np.any(~col_mask)
```
rflamary: al this below is use several time and should be a function (-one fo removing zerosif needed and one for reconstrcting the variables afterward)

ot/lp/_network_simplex.py
```diff
@@ -379,19 +379,51 @@ def emd(
...
+                alpha_init = alpha_init[row_mask]
+                beta_init = beta_init[col_mask]
 
         # Dense solver
         G, cost, u, v, result_code = emd_c(
-            a, b, M, numItermax, numThreads, alpha_init, beta_init
+            a_solver, b_solver, M_solver, numItermax, numThreads, alpha_init, beta_init
         )
 
+        if need_filter:
```
rflamary: second function function

ot/lp/_network_simplex.py
```diff
@@ -678,19 +710,57 @@ def f(b):
                 emd_c_sparse(a, b, edge_sources, edge_targets, edge_costs, numItermax)
             )
         else:
+            row_mask = asel
```
rflamary: same herre

ot/lp/_network_simplex.py
```diff
@@ -678,19 +710,57 @@ def f(b):
...
+                a_solver,
+                b_solver,
+                M_solver,
+                numItermax,
+                numThreads,
+                alpha_init,
+                beta_init,
             )
 
+            if need_filter:
```
rflamary: same here

RELEASES.md
```diff
@@ -8,6 +8,7 @@ This new release adds support for sparse cost matrices and a new lazy EMD solver
 - Add lazy EMD solver with on-the-fly distance computation from coordinates (PR #788)
 - Add Warmstart feature to the EMD solver for existing potentials (PR #793)
 - Add Warmstart potentials feature to the EMD solver for lazy and sparse solver
+- Faster init and result retrieval for EMD solver
```
rflamary: add PR number here and above

ot/lp/network_simplex_simple.h
```diff
@@ -480,13 +490,16 @@ namespace lemon {
 
             // Get cost for an arc (either from pre-computed array or compute lazily)
             inline Cost getCost(ArcsType e) const {
-                if (!_ns._lazy_cost) {
+                if (_ns._dense_cost) {
```
EduardoGRocha, line 493: I am struggling to understand the rationale for these changes. The dense cost seems to be used if the source/target vectors have no entries with 0 supply/demand. In that case, a step (`costMap`) will write a compact version of the cost matrix into `_cost`.

From the point of view of the solver, the cost matrix (whether `_cost` or `D_ptr`) is always dense. So I would argue that this compression step should be factored out of the class.

Now the simplex class carries too much internal state that makes refactorings more difficult

EduardoGRocha: To a certain extend, I feel like the arc mixing logic suffers from a similar problem. The mixing should happen before passing the input to the solver. Otherwise the solver needs to track too many possible internal states

Issue #807, opened 2026-04-29
Add `normalize` parameter to sliced Wasserstein distance functions


Harguna:
## 🚀 Feature
Add an optional `normalize` parameter (with companion `normalize_mode`) to `ot.sliced_wasserstein_distance` and `ot.max_sliced_wasserstein_distance` that standardizes inputs before computing the distance. Supports `'standard'`, `'minmax'`, and `'l2'` normalization, with `'joint'`, `'source'`, or `'target'` as the reference for fitting statistics. Default is `normalize=None`, preserving fully backward-compatible behavior.

### Motivation
Sliced Wasserstein Distance projects distributions onto random 1D directions and averages the 1D Wasserstein distances. Because projection is a dot product with a random unit vector, **features with larger numerical ranges dominate the projection**, regardless of their actual importance.

Example: comparing two user populations on `annual income` ($20k–$500k) and `interaction rate` (0–1). If incomes are similar between the populations but interaction rates differ meaningfully, the projection will still be dominated by income and SWD reports a small distance, missing the real signal.

Two compounding problems:
1. **Users often won't realize this is happening:** Unlike Euclidean distance, SWD's scale-sensitivity isn't obvious from its definition. Users may interpret a small SWD as similarity, not realizing one feature drowned out the rest.

2. **Manual normalization is friction and easy to get wrong:** Today, users must add boilerplate before every call which adds additional steps:
   ```python
   scaler = StandardScaler().fit(np.concatenate([X_s, X_t]))
   swd = ot.sliced_wasserstein_distance(scaler.transform(X_s), scaler.transform(X_t))
   ```

3. **Independent normalization:** A common mistake - fitting each distribution independently, which silently corrupts the distance with no warning:
   ```python
   X_s_norm = StandardScaler().fit_transform(X_s)
   X_t_norm = StandardScaler().fit_transform(X_t)  # WRONG: independent fit
   ```
   This makes genuinely different distributions appear identical because each is independently re-centered to mean zero.

### Pitch
Add two optional keyword arguments to both `sliced_wasserstein_distance` and `max_sliced_wasserstein_distance`:

```python
ot.sliced_wasserstein_distance(
    X_s, X_t,
    ...,
    normalize=None,           # None | 'standard' | 'minmax' | 'l2'
    normalize_mode='joint',   # 'joint' | 'source' | 'target'
)
```

`normalize` controls the method:
- `None` — no normalization (default, identical to current behavior)
[...]

PR #808, opened 2026-04-29
[MRG] Add normalize parameter to sliced_wasserstein_distance

Harguna:
## Types of changes

- [x] New Feature

Addresses #807.

Sliced Wasserstein Distance is sensitive to feature scale: features with larger numerical ranges dominate the random projections, drowning out meaningful differences in smaller-scale features. Users often don't realize this is happening and, when they do, the manual fix (preprocessing inputs with a scaler) is verbose and easy to get wrong — fitting each distribution independently silently corrupts the distance.

This PR adds optional `normalize` and `normalize_mode` parameters to `sliced_wasserstein_distance` and `max_sliced_wasserstein_distance` to handle this cleanly inside the function. Default behavior (`normalize=None`) is unchanged, so the change is fully backward-compatible.

**This is a `[WIP]` skeleton PR** - it establishes the API surface, signatures, docstrings, and a helper function so the design can be reviewed before the full implementation lands. The actual normalization math, edge case handling, behavioral tests, and example script will follow in subsequent commits on this same branch.

In this skeleton:
- Existing `test/test_sliced.py` test suite continues to pass (verifies the new keyword parameters didn't break anything).
- `pre-commit run --all-files` passes locally.

Tests related to the new feature will be added with the full implementation in the subsequent commits.

## PR checklist


rflamary:
Thanks for this PR, we are a bit busy at te moement and will have more time to give some feedback after the neurips deadline in two weeks.

Harguna:
Sounds good, best of luck with the NeurIPS.

rflamary:
Hello @Harguna , thanks for the PR.

We had a look with @clbonet and we are not really comfortable with having a normalization inside the sliced wasserstein function. While this might make sens in some applications it also means that for instance when optimizing the SWD, the loss between two optimization steps or minibtach is not comparable (since normalized locally) which poses a practical problem because it is an intuitive behavior and leads to different minimizers.

But we agree with you that normalization should be easier to handle. So we propose to handle it in a slightly different way as follows :

```python
scaler = ot.utils.DataScaler(norm='standard').fit([X_s,X_t]) # can take a tensor or a list for joint normalization
swd = ot.sliced_wasserstein_distance(X_s, X_t, scaler=scaler)
```
this means that the normalization is fitted outside on a class (compatible with sklearn with a fit and transform function but that handles backends). The `scaler` parameter should also accept a function (detected with `__call__` and can apply it so this would allow pytorch pre-processing pipeline or models). I thinks we need a helper function
```python
def apply_scaler(X_s, X_t, scaler=None)
```
 that handles the preprocessing of the data (or not if scaler=None) so that we can add this API to other functions in POT such as `ot.solve_sample`.

Would you be OK with implementing our suggestions?

Harguna:
Hello @rflamary, 

Thanks for the detailed feedback, this makes sense. I had accounted for the relative shift between `X_s` and `X_t` between optimization steps for the same batch by fitting normalization statistics jointly on `concat(X_s, X_t)`, but you're right that this doesn't address inter-batch variations, which would destabilize the objective during gradient-based training.

I agree with your suggested design which decouples the fitting step from the distance computation. I'm happy to implement DataScaler with backend compatibility and apply_scaler as a standalone helper so it can be reused across other POT functions. I will get started on that.

Harguna:
Hello @rflamary, 

I've implemented the changes based on your feedback. The `DataScaler` class and `apply_scaler` helper have been added to `ot/utils.py`, and both `sliced_wasserstein_distance` and `max_sliced_wasserstein_distance` now accept a scaler parameter. Tests are included for all new functionality. Happy to update the documentation and example gallery once you've had a chance to review.

Thanks,
Harguna

ot/utils.py
```diff
@@ -1493,6 +1493,222 @@ def check_number_threads(numThreads):
...
+        X_scaled : array-like
+            Transformed data, same shape and backend as X.
+        """
+        if self.norm != "l2" and not self._fitted:
+            raise RuntimeError(
+                "DataScaler must be fitted before calling transform() "
+                "for norm='{}'.".format(self.norm)
+            )
+
+        nx = get_backend(X)
```
rflamary, line 1635: you should test thta the nx backend is the same as sotred in seld._nx or else raise an error

ot/utils.py
```diff
@@ -1493,6 +1493,222 @@ def check_number_threads(numThreads):
...
+        X : array-like or list of array-like
+
+        Returns
+        -------
+        X_scaled : array-like or list of array-like
+            If X was a list, returns a list of transformed arrays.
+        """
+        self.fit(X)
+        if isinstance(X, (list, tuple)):
+            return [self.transform(x) for x in X]
```
rflamary: i'm OK with hnadling lists of array in fit_transform but it should also work on transform. maybe creata a sel.apply_transform_ function with all teh if/else than can be called in both transform and fit transform?

test/test_sliced.py
```diff
@@ -697,3 +697,68 @@ def test_linear_sliced_sphere_backend_type_devices(nx):
 
         nx.assert_same_dtype_device(xb, valb)
         np.testing.assert_almost_equal(sw_np, nx.to_numpy(valb))
+
+
+class TestSlicedWassersteinScaler:
+    """Integration tests for the scaler parameter in sliced_wasserstein_distance."""
+
+    def test_scaler_none_matches_no_scaler(self):
```
rflamary, line 705: please add a test that tests for different  beckends and checks that the fit and transform works accross them

see how it is done in teh following test function with parameter nx doing the loop automatically:
[def test_sliced_backend(nx):](https://github.com/PythonOT/POT/blob/19db3ee4745872d6a2380f08de13133e3dd847a9/test/test_sliced.py#L129)

also do teh same for the DataScaler class

rflamary (commented):
Hello @Harguna thanks for the modifications.

We are nearly there I just need a little more consistence between transform and fit_transform so that they can both accept lists. Also we need test of the new classes across backends . See comments below.

test/test_sliced.py
Harguna: Addressed

ot/utils.py
Harguna: Addressed

ot/utils.py
Harguna: Addressed

Harguna:
Hey @rflamary, 

Thanks for the review and for catching these gaps. I've addressed your comments and made a few minor improvements. Happy to iterate further if anything needs changing.

Best, 
Harguna

PR #833, opened 2026-07-06
[MRG] Fix label-aware cost correction in ot.da (closes #664)

### LINKED ISSUE #664: Incorrect computation of `cost_correction` matrix in `ot.da.EMDTransport`
## Describe the bug
It seems that the `cost_correction` matrix is computed incorrectly. This is the current code that can be found [here](https://github.com/PythonOT/POT/blob/47c592584d127ee01d2fa93a55cc440146fba82b/ot/da.py#L509-L518):

```python
# labels_match is a (ns, nt) matrix of {True, False} such that
# the cells (i, j) has False if ys[i] != yt[i]
label_match = (ys[:, None] - yt[None, :]) != 0 
# cost correction is a (ns, nt) matrix of {-Inf, float, Inf} such
# that he cells (i, j) has -Inf where there's no correction necessary
# by 'correction' we mean setting cost to a large value when
# labels do not match
# ...
with warnings.catch_warnings():
    warnings.simplefilter('ignore', category=RuntimeWarning)
    cost_correction = label_match * missing_labels * self.limit_max
```
The issues:
- First, the comment says that `label_match` is False if `ys[i] != yt[i]`.
However, if `ys[i] != yt[i]` then `(ys[:, None] - yt[None, :]) != 0` will be True, hence `label_match` will be True - although the labels do not match (the naming is confusing in this case). Therefore, either
  - the variable should be named `label_mismatch` and the comment should be fixed OR    
  - we check for equality `label_match = (ys[:, None] - yt[None, :]) == 0` and flip the value in `cost_correction`, i.e. `cost_correction = (1 - label_match) * ...`
- Second, `cost_correction = label_match * missing_labels * self.limit_max` will apply a cost correction only if `missing_labels` is True. However, it must not correct if `missing_labels` is True - hence, we need to flip it to `... * (1 - missing_labels ) * ...`

Therefore, I'd propose the following change

```python
# label_mismatch is a (ns, nt) matrix of {True, False} such that
# the cells (i, j) has True if ys[i] != yt[i]
label_mismatch = (ys[:, None] - yt[None, :]) != 0 
# cost correction is a (ns, nt) matrix of {-Inf, float, Inf} such
# that he cells (i, j) has -Inf where there's no correction necessary
# by 'correction' we mean setting cost to a large value when
# labels do not match
# ...
with warnings.catch_warnings():
    warnings.simplefilter('ignore', category=RuntimeWarning)
    cost_correction = label_mismatch * (1 - missing_labels) * self.limit_max
```
Happy to send the corresponding PR if you agree.

#### Screenshots
The following screenshots show the effect of flipping the `missing_labels` value. Here we map samples across multiple Gaussian distributions with 2 labels (p = 1 and p = 2). All labels are given. Without the fix, the transport plans are not computed correctly. With the fix, only samples from the same target class are linked.  

image

image

Linux-4.18.0-372.75.1.el8_6.x86_64-x86_64-with-glibc2.28
Python 3.12.4 | packaged by conda-forge | (main, Jun 17 2024, 10:23:07) [GCC 12.3.0]
NumPy 2.0.0
SciPy 1.14.0
POT 0.9.4 (pip installed)

CodingSelim:
## Types of changes

Closes #664.

The cost correction in `BaseTransport.fit` that is supposed to push apart source and target samples with different labels was computed backwards. It built `missing_ys = (ys == -1)` / `missing_yt = (yt == -1)` and used their outer product `missing_labels`, so the large `limit_max` cost was applied only where *both* labels are missing, and `label_match` was actually a mismatch mask. The net effect:

- when all labels are known (`missing_labels` is all zeros) the correction is a complete no-op, so `ys`/`yt` had no influence on the transport at all
- it also contradicted the code`s own comment ("0 iff either masked").

This is why a supervised fit gave the same cost matrix as an unsupervised one.

Fix: use present masks `(ys != -1)` / `(yt != -1)`, so the correction applies exactly to labeled source/target pairs whose labels differ. This matches the original pre-vectorized loop (the `for c in classes: cost_[idx_s, idx_t] = limit_max` version) that the vectorization in #587 replaced.

## How has this been tested

- The semisupervised checks in `test_da.py` had been flipped when the bug was introduced to assert `n_unsup == n_semisup` (i.e. that supervised mode does nothing), while the comment right above still said the norms should be different. Restored them to assert the cost actually changes.
- Added `test_semisupervised_cost_correction`: with all labels known, different-label pairs get pushed to `limit_max` and same-label pairs keep their smaller cost; in the semisupervised case, columns with an unlabeled (-1) target are never forbidden. It fails on current `master` and passes with this change.
- Full `test_da.py` passes locally.

## PR checklist

- [x] I have read the CONTRIBUTING document.
- [x] All tests passed, and additional code has been covered with new tests.
- [x] I have added the PR and Issue fix to the RELEASES.md file.

### DIFF
--- a/ot/da.py
+++ b/ot/da.py
@@ -48,6 +48,9 @@ from .mapping import (
     joint_OT_mapping_kernel,
 )
 
+# value used in ys/yt to mark a sample whose label is unknown
+MISSING_LABEL = -1
+
 
 def sinkhorn_lpl1_mm(
     a,
@@ -600,9 +603,9 @@ class BaseTransport(BaseEstimator):
 
                 # present_labels is a (ns, nt) matrix of {0, 1} such that
                 # the cell (i, j) is 1 iff both ys[i] and yt[j] are labeled
-                # (i.e. neither is masked with -1)
-                present_ys = (ys != -1) + nx.zeros(ys.shape, type_as=ys)
-                present_yt = (yt != -1) + nx.zeros(yt.shape, type_as=yt)
+                # (i.e. neither is masked with MISSING_LABEL)
+                present_ys = (ys != MISSING_LABEL) + nx.zeros(ys.shape, type_as=ys)
+                present_yt = (yt != MISSING_LABEL) + nx.zeros(yt.shape, type_as=yt)
                 present_labels = present_ys[:, None] @ present_yt[None, :]
                 # label_mismatch is a (ns, nt) matrix of {True, False} such that
                 # the cell (i, j) is True if ys[i] != yt[j]
--- a/ot/da.py
+++ b/ot/da.py
@@ -598,23 +598,24 @@ class BaseTransport(BaseEstimator):
                 if self.limit_max != np.inf:
                     self.limit_max = self.limit_max * nx.max(self.cost_)
 
-                # missing_labels is a (ns, nt) matrix of {0, 1} such that
-                # the cells (i, j) has 0 iff either ys[i] or yt[j] is masked
-                missing_ys = (ys == -1) + nx.zeros(ys.shape, type_as=ys)
-                missing_yt = (yt == -1) + nx.zeros(yt.shape, type_as=yt)
-                missing_labels = missing_ys[:, None] @ missing_yt[None, :]
-                # labels_match is a (ns, nt) matrix of {True, False} such that
-                # the cells (i, j) has False if ys[i] != yt[i]
-                label_match = (ys[:, None] - yt[None, :]) != 0
+                # present_labels is a (ns, nt) matrix of {0, 1} such that
+                # the cell (i, j) is 1 iff both ys[i] and yt[j] are labeled
+                # (i.e. neither is masked with -1)
+                present_ys = (ys != -1) + nx.zeros(ys.shape, type_as=ys)
+                present_yt = (yt != -1) + nx.zeros(yt.shape, type_as=yt)
+                present_labels = present_ys[:, None] @ present_yt[None, :]
+                # label_mismatch is a (ns, nt) matrix of {True, False} such that
+                # the cell (i, j) is True if ys[i] != yt[j]
+                label_mismatch = (ys[:, None] - yt[None, :]) != 0
                 # cost correction is a (ns, nt) matrix of {-Inf, float, Inf} such
                 # that he cells (i, j) has -Inf where there's no correction necessary
-                # by 'correction' we mean setting cost to a large value when
-                # labels do not match
+                # by 'correction' we mean setting cost to a large value when two
+                # labeled samples have different labels
                 # we suppress potential RuntimeWarning caused by Inf multiplication
                 # (as we explicitly cover potential NANs later)
                 with warnings.catch_warnings():
                     warnings.simplefilter("ignore", category=RuntimeWarning)
-                    cost_correction = label_match * missing_labels * self.limit_max
+                    cost_correction = label_mismatch * present_labels * self.limit_max
                 # this operation is necessary because 0 * Inf = NAN
                 # thus is irrelevant when limit_max is finite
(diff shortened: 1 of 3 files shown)


rflamary:
hello @CodingSelim the PR is looking good, what you coded makes sens to me. You need to skip the tf backend for the DA class.

@kachayev i'm open to your opinion on this too. I would like to do a release shortly

test/test_da.py
```diff
@@ -586,6 +586,38 @@ def test_emd_transport_class(nx):
     )
 
 
+def test_semisupervised_cost_correction(nx):
```
rflamary, line 590:
```suggestion
@pytest.skip_backend("tf")
def test_semisupervised_cost_correction(nx):
```

ot/da.py
```diff
@@ -598,23 +598,24 @@ class label
...
-                missing_ys = (ys == -1) + nx.zeros(ys.shape, type_as=ys)
-                missing_yt = (yt == -1) + nx.zeros(yt.shape, type_as=yt)
-                missing_labels = missing_ys[:, None] @ missing_yt[None, :]
-                # labels_match is a (ns, nt) matrix of {True, False} such that
-                # the cells (i, j) has False if ys[i] != yt[i]
-                label_match = (ys[:, None] - yt[None, :]) != 0
+                # present_labels is a (ns, nt) matrix of {0, 1} such that
+                # the cell (i, j) is 1 iff both ys[i] and yt[j] are labeled
+                # (i.e. neither is masked with -1)
+                present_ys = (ys != -1) + nx.zeros(ys.shape, type_as=ys)
```
kachayev: Minor suggestion: maybe define module-level constant, say `MASK_LABEL=-1` or `MISSING_LABEL=-1` or something similar. Just to avoid "magical" numerical constants in the code.

ot/da.py
```diff
@@ -598,23 +598,24 @@ class label
...
+                label_mismatch = (ys[:, None] - yt[None, :]) != 0
                 # cost correction is a (ns, nt) matrix of {-Inf, float, Inf} such
                 # that he cells (i, j) has -Inf where there's no correction necessary
-                # by 'correction' we mean setting cost to a large value when
-                # labels do not match
+                # by 'correction' we mean setting cost to a large value when two
+                # labeled samples have different labels
                 # we suppress potential RuntimeWarning caused by Inf multiplication
                 # (as we explicitly cover potential NANs later)
                 with warnings.catch_warnings():
```
kachayev, line 619: I wonder if the same logic could be replicated without tempering with warnings propagation.... something like 

```
correction_mask = label_mismatch & present_labels
cost_correction = nx.full_like(self.cost_, -np.inf)
cost_correction[correction_mask] = self.limit_max
self.cost_ = nx.maximum(self.cost_, cost_correction)
```

I'm not sure about the code above, didn't have a chance to properly test it!

kachayev:
@rflamary Looks good to me! Left a couple of minor comments/suggestions.

CodingSelim:
thanks for the review! pushed the MISSING_LABEL constant.

on the cost_correction refactor, i think the item assignment (cost_correction[mask] = self.limit_max) won't work across backends, jax arrays are immutable and the tf tensor path doesn't support that kind of in-place scatter either, which is why the original avoids indexing and does it with arithmetic. i kept the multiply approach but it should be fine, the only reason for the catch_warnings is the 0*inf case when limit_max is left at the default inf, and we already scrub those with nan_to_num right after. happy to try a where-based version if you'd prefer that over the warnings filter, that one stays backend safe.

rflamary:
Yep jax is a pain for generic backends... I think the PR looks great, let us wait for tetss to pass and merge it

kachayev:
@CodingSelim you are right! don't worry, it's a chunk of old code anyways

PR #838, opened 2026-08-25
[MRG] Add QSW sampling for sliced Wasserstein

### LINKED ISSUE #835: [Feature Request] Quasi-Monte Carlo point sets for the Sliced Wasserstein module
## Motivation

`ot/sliced.py` currently generates projection directions purely at random
(`get_random_projections` normalizes Gaussian vectors; `get_projections_sphere`
uses QR on a random matrix) — standard Monte Carlo (MC) sampling over the
sphere.

A recent paper, ["Quasi-Monte Carlo for 3D Sliced Wasserstein"](https://arxiv.org/abs/2309.11713)
(Nguyen, Bariletto & Ho, ICLR 2024), shows that replacing these random
directions with deterministic, low-discrepancy **QMC point sets** on the
sphere reduces the approximation error of Sliced Wasserstein, especially in
3D. They call this **Quasi-Sliced Wasserstein (QSW)**.

Since QSW is deterministic, the paper also proposes **Randomized QSW
(RQSW)**: applying a random rotation to a QMC point set, which provably
preserves both low discrepancy and unbiasedness — making it a drop-in,
unbiased replacement for the current random sampling in stochastic
optimization settings (e.g. gradient flows).

## What I'd like to propose

A scoped-down first PR covering one construction:

- A QMC point set on the sphere via **generalized spiral points** (simple,
  closed-form, no optimization needed).
- Its randomized version, **RQSW via random rotation** — reusing the same
  QR-based rotation mechanism `get_projections_sphere` already has, just for
  a full `(d, d)` rotation instead of `(d, 2)`.

The paper covers other QMC constructions too (equal-area mapping,
energy-minimizing points) as possible follow-ups if this is of interest.

## Before I start

1. Would a QMC-based option for this module be of interest?
2. New dedicated function, or a `method=` option on the existing API?

Happy to adjust based on feedback. Thanks!

Samuel-Vangu:
## Types of changes

* [x] New feature
* [x] Documentation update
* [x] Tests

Closes #835

This PR adds Quasi-Monte Carlo (QMC) sampling of projection directions to the Sliced Wasserstein module.

Currently, `sliced_wasserstein_distance` samples projection directions uniformly at random, corresponding to standard Monte Carlo sampling. This PR adds two alternatives based on the generalized spiral point construction described in [[Nguyen, Bariletto & Ho (2024)](https://arxiv.org/abs/2309.11713)](https://arxiv.org/abs/2309.11713):

* `sampling_slices="qsw"`: deterministic Quasi-Sliced Wasserstein (QSW) projection directions.
* `sampling_slices="rqsw"`: Randomized QSW (RQSW), obtained by applying a random rotation to the deterministic spiral point set.

The new sampling methods are currently limited to 3D, while the existing `"uniform"` sampling remains the default.

The implementation also exposes `get_projections_spiral` and updates the documentation, README references, release notes, and adds a 3D example.

The changes have been tested with:

* `pre-commit run --all-files` — all checks pass.
* `pytest test/sliced/test_sliced_distances.py` — **62 tests passed**.
* Added tests covering:

  * deterministic spiral projections;
  * randomized projections and sphere preservation;
  * seed reproducibility;
  * invalid dimensions and sampling methods;
  * QSW/RQSW with NumPy, JAX, PyTorch and TensorFlow backends;
  * consistency of deterministic QSW across backends;
  * QSW approximation compared with uniform Monte Carlo sampling in 3D.

## PR checklist

* [x] I have read the [**[CONTRIBUTING](https://pythonot.github.io/contributing.html)**](https://pythonot.github.io/contributing.html) document.
* [x] The documentation is up-to-date with the changes I made (check build artifacts).
* [x] All tests passed, and additional code has been **covered with new tests**.
* [x] I have added the PR and Issue fix to the [**RELEASES.md**](RELEASES.md) file.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,4 +1,13 @@
 # Releases
+## 0.9.8dev
+*August 2026*
+
+#### New features
+
+- Add Quasi-Monte Carlo sliced Wasserstein sampling (QSW/RQSW) via generalized
+  spiral points, selectable with `sampling_slices` in `sliced_wasserstein_distance`,
+  as described in [93] (PR #xxx)
+
 
 ## 0.9.7.post1
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,7 +6,7 @@
 
 - Add Quasi-Monte Carlo sliced Wasserstein sampling (QSW/RQSW) via generalized
   spiral points, selectable with `sampling_slices` in `sliced_wasserstein_distance`,
-  as described in [93] (PR #xxx)
+  as described in [93] (PR #838)
 
 
 ## 0.9.7.post1
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,19 +1,13 @@
 # Releases
-## 0.9.8dev
-*August 2026*
-
-#### New features
-
-- Add Quasi-Monte Carlo sliced Wasserstein sampling (QSW/RQSW) via generalized
-  spiral points, selectable with `sampling_slices` in `sliced_wasserstein_distance`,
-  as described in [93] (PR #838)
-
 
 ## 0.9.8dev
 
 #### New features
 
 - Add stereographic spherical sliced Wasserstein distance in `ot.sliced.stereographic_sliced_wasserstein_sphere`, with its rotationally invariant extension (PR #836)
+- Add Quasi-Monte Carlo sliced Wasserstein sampling (QSW/RQSW) via generalized
+  spiral points, selectable with `sampling_slices` in `sliced_wasserstein_distance`,
+  as described in [95] (PR #838)
 
 #### Closed issues
 
--- a/examples/sliced-wasserstein/plot_qsw_3d.py
+++ b/examples/sliced-wasserstein/plot_qsw_3d.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+"""
+=========================================================
+Quasi-Monte Carlo Sliced Wasserstein in 3D
+=========================================================
+
+This example illustrates the Quasi-Sliced Wasserstein (QSW) and Randomized
+Quasi-Sliced Wasserstein (RQSW) sampling schemes introduced in [93], and
+compares them to the default uniform (Monte Carlo) sampling of slicing
+directions.
+
+Sliced Wasserstein (SWD) approximates the Wasserstein distance by averaging
+1D Wasserstein distances over projections onto random directions
+:math:`\\theta` drawn uniformly on the sphere. By default these directions
+are sampled purely at random (Monte Carlo), which introduces some variance
+in the estimate for a given number of projections.
+
+QSW replaces the random directions with a deterministic, low-discrepancy
+point set on the sphere (generalized spiral points), which covers the
+sphere more evenly than random sampling and reduces the approximation
+error, especially in 3D. Since QSW is deterministic it cannot directly be
+used as an unbiased estimator in stochastic settings (e.g. gradient-based
+optimization) -- RQSW addresses this by applying a random rotation to the
+same point set, which preserves both its low discrepancy and its
+unbiasedness.
+
+We first visualize the three sampling schemes on the sphere, then measure
+how fast each one converges to a reference Sliced Wasserstein distance
+between two 3D distributions.
+
+.. [93] Nguyen, K., Bariletto, N., & Ho, N. (2024). Quasi-Monte Carlo for
+    3D Sliced Wasserstein. International Conference on Learning
+    Representations (ICLR).
+"""
+
+# Author: Samuel Vangu <samuelvangu0@gmail.com>
+#
+# License: MIT License
+
+# sphinx_gallery_thumbnail_number = 1
+
+import numpy as np
+import matplotlib.pylab as pl
+from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 (registers the 3D projection)
+
+import ot
+from ot.sliced import get_random_projections, get_projections_spiral
+
+##############################################################################
+# Visualize the three sampling schemes on the sphere
+# ----------------------------------------------------
+# We draw ~1000 directions on :math:`S^2` with each scheme:
+#
+# - ``uniform``: directions are Gaussian vectors normalized to unit norm
+#   (standard Monte Carlo sampling of the sphere).
+# - ``qsw``: deterministic generalized spiral points -- a simple,
+#   closed-form low-discrepancy point set (Rakhmanov, Saff & Zhou, 1994).
+#   The same call always returns the same points.
+# - ``rqsw``: the same spiral point set, rotated by a random (3, 3)
+#   rotation matrix (drawn via QR decomposition of a Gaussian matrix).
+#   The rotation makes the estimator unbiased while keeping the points
+#   as evenly spread out as the deterministic QSW set.
+
+n_projections = 1000
+d = 3
+seed = 42
+
+theta_uniform = get_random_projections(d, n_projections, seed=seed)
+theta_qsw = get_projections_spiral(d, n_projections, randomized=False)
+theta_rqsw = get_projections_spiral(d, n_projections, randomized=True, seed=seed)
+
+fig = pl.figure(1, figsize=(15, 5))
+
+schemes = [
+    (theta_uniform, "Uniform (Monte Carlo)"),
+    (theta_qsw, "QSW (deterministic spiral)"),
+    (theta_rqsw, "RQSW (randomly rotated spiral)"),
+]
+
+for i, (theta, title) in enumerate(schemes):
+    ax = fig.add_subplot(1, 3, i + 1, projection="3d")
+    ax.scatter(theta[0], theta[1], theta[2], c=theta[2], cmap="viridis", s=4, alpha=0.8)
+    ax.set_title(title)
+    ax.set_box_aspect([1, 1, 1])
+    ax.view_init(elev=20, azim=45)
+    ax.set_xticks([])
+    ax.set_yticks([])
+    ax.set_zticks([])
+
+pl.tight_layout()
+pl.show()
+
+# Notice how the uniform sample leaves visible gaps and clusters, while QSW
+# and RQSW spread the points much more evenly over the sphere -- this is
+# exactly the low-discrepancy property that reduces the error of the Sliced
+# Wasserstein estimate.
+
+##############################################################################
+# Convergence to the true Sliced Wasserstein distance
+# ------------------------------------------------------
+# We now compare how fast each sampling scheme converges to a reference SWD
+# value as the number of projections grows, on two 3D Gaussian samples.
+# For a fixed number of projections, ``uniform`` and ``rqsw`` are random
+# estimators (we average their error over several trials), while ``qsw`` is
+# deterministic (a single value per number of projections).
+
+rng = np.random.RandomState(0)
+
+n_samples = 200
+Xs = rng.multivariate_normal([0, 0, 0], np.eye(3), n_samples)
+Xt = rng.multivariate_normal([1.5, 1.0, -0.5], np.eye(3), n_samples)
+
+# "Ground truth" SWD, estimated with a very large number of uniform
+# projections so that the residual Monte-Carlo error is negligible.
+sw_true = ot.sliced_wasserstein_distance(Xs, Xt, n_projections=200000, seed=0)
+
+n_proj_list = [10, 20, 50, 100, 200, 500, 1000]
+n_trials = 20
+
+errors_uniform = np.zeros((n_trials, len(n_proj_list)))
+errors_rqsw = np.zeros((n_trials, len(n_proj_list)))
+errors_qsw = np.zeros(len(n_proj_list))
+
+for j, n_proj in enumerate(n_proj_list):
+    for t in range(n_trials):
+        sw_uniform = ot.sliced_wasserstein_distance(
+            Xs, Xt, n_projections=n_proj, sampling_slices="uniform", seed=t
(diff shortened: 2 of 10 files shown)


Samuel-Vangu:
Hi @rflamary, @clbonet,

This one is ready for your review whenever you have time. Thanks!

ot/sliced/_sliced_distances.py
```diff
@@ -38,6 +39,12 @@ def sliced_wasserstein_distance(
 
     - :math:`\theta_\# \mu` stands for the pushforwards of the projection :math:`X \in \mathbb{R}^d \mapsto \langle \theta, X \rangle`
 
+    By default, the projection directions :math:`\theta` are sampled uniformly
+    at random. Setting ``sampling_slices`` to ``"qsw"`` or ``"rqsw"`` instead
+    uses Quasi-Monte Carlo point sets on the sphere (generalized spiral
```
clbonet, line 44: Is the generalized spiral the best QMC method? With the current API, we can't really add other QMC methods such as Equal area mappings, Coulomb energy etc...

Would it be worth it to add this in the names?

ot/sliced/_utils.py
```diff
@@ -235,6 +235,87 @@ def projection_sphere_to_circle(
     return Xp_coords, projections
 
 
+def get_projections_spiral(
+    d, n_projections, randomized=True, seed=None, backend=None, type_as=None
+):
+    r"""
+    Generates n_projections points on the sphere via generalized
+    spiral points (Rakhmanov, Saff & Zhou, 1994).
```
clbonet: Please add the ref

ot/sliced/_utils.py
```diff
@@ -235,6 +235,87 @@ def projection_sphere_to_circle(
     return Xp_coords, projections
 
 
+def get_projections_spiral(
+    d, n_projections, randomized=True, seed=None, backend=None, type_as=None
```
clbonet, line 239: Generalized spirals are only valid for d==3, so maybe we don't need d here

ot/sliced/_utils.py
```diff
@@ -235,6 +235,87 @@ def projection_sphere_to_circle(
...
+    True
+    >>> rprojs = get_projections_spiral(d, n_projections, randomized=True, seed=0)
+    >>> np.allclose(np.sum(np.square(rprojs), 0), 1.)  # doctest: +NORMALIZE_WHITESPACE
+    True
+
+    """
+    if d != 3:
+        raise ValueError(
+            f"get_projections_spiral is only implemented for d=3, got d={d}"
+        )
```
clbonet, line 290: Not needed?

ot/sliced/_sliced_distances.py
```diff
@@ -120,9 +143,23 @@ def sliced_wasserstein_distance(
...
+                d, n_projections, seed, backend=nx, type_as=X_s
+            )
+        elif sampling_slices == "qsw":
+            projections = get_projections_spiral(
+                d, n_projections, randomized=False, backend=nx, type_as=X_s
+            )
+        elif sampling_slices == "rqsw":
+            projections = get_projections_spiral(
+                d, n_projections, randomized=True, seed=seed, backend=nx, type_as=X_s
+            )
```
clbonet: Maybe put "randomized_qsw" in the name, and then use a variable `randomized="randomized" in sampling_slices`. It would be better if we want to add more QMC sampling methods later

ot/sliced/_utils.py
```diff
@@ -235,6 +235,87 @@ def projection_sphere_to_circle(
...
+
+    if backend is None:
+        nx = NumpyBackend()
+    else:
+        nx = backend
+
+    # sin/cos are not exposed by the backend abstraction (only arccos, atan2
+    # exist), so the deterministic point construction is done in plain NumPy
+    # and converted to the target backend at the end.
```
clbonet: You can add sin and cos to the backend and use nx instead of np

RELEASES.md
```diff
@@ -1,4 +1,13 @@
 # Releases
+## 0.9.8dev
+*August 2026*
+
+#### New features
```
clbonet: To delete

RELEASES.md
```diff
@@ -1,4 +1,13 @@
...
+
+- Add Quasi-Monte Carlo sliced Wasserstein sampling (QSW/RQSW) via generalized
+  spiral points, selectable with `sampling_slices` in `sliced_wasserstein_distance`,
+  as described in [93] (PR #838)
```
clbonet: To move below

clbonet (commented):
Thank you @Samuel-Vangu for the great PR! Overall the code is great.

About the name of the method in "sampling_slices", I am wondering whether we should put something more precised than "qsw". The generalized_spiral seems to be the fastest method, but maybe we would like to add other Quasi Monte-Carlo methods in the future such as the minimization of the Coulomb energy or something else. Thus, I think we should put something more precised, e.g. "spiral_qmc"? 

Also, for the randomized option. Since it can be applied to any Quasi-Monte Carlo method, maybe it should be "randomized_spiral_qmc", and we can get the boolean with checking `"randomized" in sampling_slices`?

What do you think @Samuel-Vangu, @rflamary ? 

Otherwise, I have few minor comments below.

examples/sliced-wasserstein/plot_qsw_3d.py
```diff
@@ -0,0 +1,192 @@
...
+optimization) -- RQSW addresses this by applying a random rotation to the
+same point set, which preserves both its low discrepancy and its
+unbiasedness.
+
+We first visualize the three sampling schemes on the sphere, then measure
+how fast each one converges to the true Sliced Wasserstein distance
+between two point clouds -- known here in closed form, with no
+approximation error left except from the number of projections itself.
+
+.. [93] Nguyen, K., Bariletto, N., & Ho, N. (2024). Quasi-Monte Carlo for
```
clbonet: Update [93] to [95]

Samuel-Vangu:
Hi @clbonet,

Thanks for the detailed review! Here's a summary of what I changed in response to each point:

**Naming (`sampling_slices` values)**
Agreed — `"qsw"`/`"rqsw"` were too generic given that spiral is just one of several possible QMC constructions (equal-area, Coulomb energy, etc. could come later). Renamed to `"spiral_qmc"` and `"randomized_spiral_qmc"`, and the dispatcher now derives the boolean via `sampling_slices.startswith("randomized_")` plus a `method = sampling_slices.removeprefix("randomized_")`, so adding a new QMC construction later only needs one new `elif`, not two.

**Missing reference**
Added Rakhmanov, Saff & Zhou (1994) as `[96]` in the README, and cited it in `get_projections_spiral`'s docstring alongside `[95]` (Nguyen et al.) which is now correctly numbered.

**`d` parameter**
Kept it in the signature — for consistency with `get_projections_sphere` and so future d-generic constructions (Coulomb, equal-area) fit the same dispatch pattern — but added `if d != 3: raise ValueError(...)` **inside** `get_projections_spiral` itself, not just in the dispatcher. That way a user calling the function directly (not through `sliced_wasserstein_distance`) is still protected, not only callers going through `sampling_slices`.

**sin/cos on the backend**
Added `sin`/`cos` to `Backend` and all five implementations. I kept `get_projections_spiral`'s deterministic point construction in plain NumPy though, rather than switching to `nx`: most backends' `arange()` doesn't propagate `type_as`'s dtype (e.g. JAX defaults to float32 without x64 enabled), so building natively per-backend would likely break `test_qsw_matches_across_backends`'s exact cross-backend equality check. 

**RELEASES.md**
Removed the duplicate `## 0.9.8dev` header and merged the entry under the existing section.

**`[93]` → `[95]`**
Fixed in `plot_qsw_3d.py` and everywhere else it appeared.

Let me know if any of these choices don't sit right.

ot/backend.py
```diff
@@ -1225,6 +1225,26 @@ def slogdet(self, a):
...
+
+    def cos(self, a):
+        r"""
+        Trigonometric cosine, element-wise.
+
+        This function follows the api from :any:`numpy.cos`
+
+        See: https://numpy.org/doc/stable/reference/generated/numpy.cos.html
+        """
+        raise NotImplementedError()
```
clbonet, line 1246: Add tests in test/test_backend

clbonet (commented):
Thank you @Samuel-Vangu. A last small comment, if you add sin and cos to the backend, can you also please add the tests.

Samuel-Vangu:
Hi @clbonet ,

Added sin/cos coverage to both test_empty_backend and test_func_backends,
as requested. All green (test_backend.py + test_sliced_distances.py, 62
tests).

test/sliced/test_sliced_distances.py
```diff
@@ -403,3 +403,269 @@ def test_sliced_wasserstein_scaler_backend(nx):
...
+    rng = np.random.RandomState(0)
+    X_s = rng.normal(0, 1, (30, d))
+    X_t = rng.normal(1, 1, (30, d))
+
+    backends = ot.backend.get_backend_list()
+    assert len(backends) >= 1
+
+    results = {}
+    projections_by_backend = {}
+    for nx in backends:
```
rflamary: no loop inside the etsts please pass nx as a parameter of the functin and pytest will loop over available backends

test/sliced/test_sliced_distances.py
```diff
@@ -403,3 +403,269 @@ def test_sliced_wasserstein_scaler_backend(nx):
...
+        pytest.skip("Need at least two installed backends to compare")
+
+    d = 3
+    n_projections = 50
+    rng = np.random.RandomState(0)
+    X_s = rng.normal(0, 1, (30, d))
+    X_t = rng.normal(1, 1, (30, d))
+
+    results = {}
+    for nx in backends:
```
rflamary: same here

test/sliced/test_sliced_distances.py
```diff
@@ -403,3 +403,269 @@ def test_sliced_wasserstein_scaler_backend(nx):
...
+
+    val_qsw = ot.sliced_wasserstein_distance(
+        X_s, X_t, n_projections=n_projections, sampling_slices="spiral_qmc"
+    )
+    qsw_error = abs(val_qsw - reference)
+
+    mean_uniform_error = np.mean(uniform_errors)
+    mean_rqsw_error = np.mean(rqsw_errors)
+
+    print(f"\n[DEBUG] mean uniform error={mean_uniform_error:.6e}")
```
rflamary: remove the debugging print

ot/sliced/_utils.py
```diff
@@ -235,6 +235,89 @@ def projection_sphere_to_circle(
...
+    phi2 = (1.8 * np.sqrt(n_projections) * phi1) % (2 * np.pi)
+    theta_np = np.stack(
+        [np.sin(phi1) * np.cos(phi2), np.sin(phi1) * np.sin(phi2), z], axis=0
+    )  # shape (d, n_projections)
+    theta = nx.from_numpy(theta_np, type_as=type_as)
+
+    if not randomized:
+        return theta
+
+    if isinstance(seed, np.random.RandomState) and str(nx) == "numpy":
```
rflamary, line 309: this test should not be there, nx.seed and nx.randn shoudl work with  numpy backend too

rflamary (commented):
Hello @Samuel-Vangu 

Thanks for the contribution. I detected a few remeianing somments thta need sto be adressed but we are close to merging the PR.

Samuel-Vangu:
Hi @rflamary, addressed all three test/sliced/test_sliced_distances.py
comments: both loops replaced with `nx` as a function parameter (pytest
now loops over backends automatically), and the debug print removed --
the thresholds it was showing are documented in the test's docstring
instead.

On the `isinstance(seed, np.random.RandomState)` check in `_utils.py`: I
tested this before removing it, and I believe it's actually needed.
`RandomState.seed()` doesn't accept another `RandomState` instance as
input:

```python
import numpy as np
rng_self = np.random.RandomState()
rng_passed = np.random.RandomState(42)
rng_self.seed(rng_passed)
# TypeError: Cannot cast scalar from dtype('O') to dtype('int64')
#            according to the rule 'safe'
```

So the generic `nx.seed(seed)` path raises when someone passes an
already-instantiated `RandomState` rather than an int. This is also the
exact same pattern already used in `get_random_projections`,
`get_projections_sphere`, and `get_random_rotations` just above in this
same file, which I didn't touch.

rflamary:
Hello @Samuel-Vangu thanks for the fixeds.

About nx.seed : if it breaks then it means that we need to fix the numpy backend (and do the test in the nx.seed function). We need a backend that can run this without if/then tests.  Coiuld you do that please? And thank you for identifying this bug.

Samuel-Vangu:
@rflamary ,     
                                                                                                                                                                                                                           Thanks! Grep found 6 hits: the same check in all 4 functions here, plus
one in _spherical_sliced.py and one in ot/utils.py that look like
different logic, worth checking separately.

I'd rather open a follow-up PR for the actual fix -- ot/backend.py is
shared code deserving its own review, and this PR is close to merging,
so I don't want to reopen a broader cycle now. Leaving the checks
untouched here for that reason. What do you think ?

rflamary:
OK we can merge but please before open an issue with precise description of what needs to be done and where (and hopefully indeed do the PR after ;) ).

The whole point of the backend is to have shared code and this is a bug so it should be listed in Issues.

Samuel-Vangu:
@rflamary ,

I just opened the issue #848

clbonet:
@Samuel-Vangu Don't forget to add your name in the Contributors file.

Samuel-Vangu:
Hi @clbonet ,

Added my name to CONTRIBUTORS.md.

PR #846, opened 2026-08-30
[MRG] Fix device placement in expected sliced plans

cananoo:
## Summary

- preserve the input dtype and device for default weights in `sliced_plans` and `expected_sliced_plan`
- compute the expected sliced-plan cost with `sparse_ot_dist` so sparse plans do not materialize a full distance matrix
- fix `sparse_ot_dist` to apply per-pair weights in batch order
- add regression coverage for dense/sparse plans, implicit/explicit weights, and weighted sparse distances

Closes #845

## Validation

- `pre-commit run --files ot/sliced/_sliced_plans.py ot/utils.py test/sliced/test_sliced_plans.py test/test_utils.py`
- NumPy source smoke tests for dense/sparse plans and weighted sparse distances
- `python -m compileall -q ot/sliced/_sliced_plans.py ot/utils.py test/sliced/test_sliced_plans.py test/test_utils.py`

The full test suite could not be built locally because this Windows environment does not have MSVC for POT's Cython extensions. The PyTorch regression test runs on CPU or CUDA when the backend is available in CI.

PR #847, opened 2026-08-30
[MRG] Create the CUDA context only when a GPU tensor is actually used

### LINKED ISSUE #612: ot.solve uses GPU even though tensors are on CPU?
## Describe the bug
Running `ot.solve` with tensors on the CPU allows memory on the GPU (thisis documented in `get_backend_list`) but also seems to use the GPU, as the Watts are increasing. See attached screencast :
[Screencast from 08-03-2024 11:24:44.webm](https://github.com/PythonOT/POT/assets/8993218/ac6ba4b6-8529-436a-987d-9c0a661fee4e)

Is it normal?

Script
```python
import torch
import ot

n_samples = 5_000

x = torch.randn(n_samples, 2)
y = torch.randn(n_samples, 2)

a = torch.rand(n_samples)
a /= a.sum()
b = torch.rand(n_samples)
b /= b.sum()

M = ot.dist(x, y)

res = ot.solve(M, a, b, reg=0.1, reg_type="entropy")
```

DenisDrobyshev:
## Types of changes

Closes #612.

`TorchBackend.__init__` builds a CUDA generator and the CUDA entries of
`__type_list__` as soon as the backend is instantiated. Both initialise a CUDA
context, so the first POT call on CPU-only torch tensors claims device memory
and wakes the GPU — which is what the issue reports as rising power draw with
CPU tensors.

On an RTX 4060 with POT 0.9.7.post1:

```python
import torch, ot
torch.cuda.is_initialized()        # False
x = torch.randn(2000, 2)           # CPU
ot.dist(x, x)
torch.cuda.is_initialized()        # True
torch.cuda.memory_allocated()      # 1024
```

The backend is instantiated lazily, which is why this happens at the first call
rather than at `import ot`.

## Description

The CUDA generator and the CUDA half of the type list are now built on first
access instead of in `__init__`:

* `__type_list__` becomes a property. The CPU entries are built eagerly as
  before; the CUDA entries are appended the first time the list is read, so
  anything that enumerates dtypes and devices (the tests, `_bench`) sees exactly
  what it saw before.
* `rng_cuda_` becomes a property with a setter, so `seed(torch.Generator)` keeps
  working.
* `seed(int)` records the seed instead of forcing the CUDA generator into
  existence, and applies it when that generator is first created. A seed set
  before any GPU work therefore still governs it.

After the change the same script leaves CUDA untouched, and a later GPU call
initialises it as usual.

## Reproducibility

Seeded GPU sampling is unchanged. With the same seed, before and after:

```python
nx.seed(42); nx.rand(3, type_as=torch.zeros(1, device="cuda"))
# [0.61296, 0.010059, 0.398414]   identical on both
```

`len(nx.__type_list__)` is still 4 when CUDA is available.

## How has this been tested

New `test_no_cuda_context_for_cpu_only_work` in `test/test_backend.py`, a
non-regression test for #612. It asserts that CPU-only work leaves
`torch.cuda.is_initialized()` false and `memory_allocated()` at zero. The check
needs an interpreter that has not touched CUDA yet, so it runs in a subprocess,
following `test_torch_optimizer_after_tensorflow_import` in the same file. It
fails on `master` and passes here, and is skipped without CUDA.

`test/test_backend.py` is 23 passed, 1 skipped (22 passed, 1 skipped before this
PR — the new test is the difference). `pre-commit run` on the changed files is
clean.

Ran on Windows with torch 2.9.1+cu126 and an RTX 4060 Laptop GPU.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,7 +9,7 @@
 #### Closed issues
 
 - Preserve input dtype and device for expected sliced plans, avoid materializing dense distance matrices for sparse plans, and fix weighted sparse-distance ordering (PR #846, Issue #845)
-- Build the CUDA generator and the CUDA entries of `TorchBackend.__type_list__` lazily, so that using POT with CPU-only torch tensors no longer initialises a CUDA context and claims device memory (Issue #612)
+- Build the CUDA generator and the CUDA entries of `TorchBackend.__type_list__` lazily, so that using POT with CPU-only torch tensors no longer initialises a CUDA context and claims device memory (PR #847, Issue #612)
 - Fix the sign issue in updates of the previous transport plan in `ot.batch.proximal_bregman_log_plan_batch` (Issue #842)
 - Load triton before TensorFlow in `ot.backend` so that building a torch optimizer no longer segfaults the interpreter, and remove the `torch<2.12` pin from the doctest and documentation requirements (PR #839, Issue #816)
 - Fix mean centering in `ot.dr.fda` and `ot.dr.wda`: `np.mean(X)` returned a scalar instead of the per-feature mean, so `proj` did not center the data as documented. In `ot.dr.fda` the same pattern in the class means made the between-class scatter matrix independent of which features separate the classes, and FDA returned a non-discriminant direction (PR #840)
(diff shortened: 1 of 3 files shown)


DenisDrobyshev:
Rebased onto master after #846 landed. The only conflict was in `RELEASES.md`, where both PRs added an entry under *Closed issues*; both entries are now kept. The code change is untouched, and `test/test_backend.py` is 23 passed, 1 skipped on the rebased branch.

rflamary:
Thanks @DenisDrobyshev  will doa code review ASAP

rflamary (approved):

PR #849, opened 2026-09-07
[MRG] Support RandomState instances in NumpyBackend.seed

### LINKED ISSUE #848: NumpyBackend.seed() rejects a np.random.RandomState instance, forcing callers to special-case it
### Describe the bug
`NumpyBackend.seed()` only accepts an int (or `None`). Passing an
already-instantiated `np.random.RandomState` raises a `TypeError`,
because `RandomState.seed()` itself doesn't accept another `RandomState`
as input:

```
TypeError: Cannot cast scalar from dtype('O') to dtype('int64')
           according to the rule 'safe'
```

As a result, every caller that wants to support passing an existing
`RandomState` has to special-case it manually instead of relying on the
shared `nx.seed(seed)` / `nx.randn(...)` abstraction. `TorchBackend` and
`TensorflowBackend` already handle this correctly -- both adopt an
externally-passed generator (`torch.Generator`, `tf.random.Generator`)
directly. `NumpyBackend` is the outlier.

#### Code sample
```python
import numpy as np
from ot.backend import NumpyBackend

nx = NumpyBackend()
rng = np.random.RandomState(42)
nx.seed(rng)  # raises TypeError
```

### Expected behavior
`nx.seed(rng)` should adopt `rng` as the backend's own generator (same
pattern as `TorchBackend`/`TensorflowBackend`), so later `nx.randn(...)`
calls draw from it, with no error and no special-casing needed by callers.

Fix, in `NumpyBackend.seed()`:
```python
def seed(self, seed=None):
    if isinstance(seed, np.random.RandomState):
        self.rng_ = seed
    elif seed is not None:
        self.rng_.seed(seed)
```

Once this lands, the `isinstance(seed, np.random.RandomState) and str(nx)
== "numpy"` special case can be removed from 4 places in
`ot/sliced/_utils.py` (`get_random_projections`, `get_projections_sphere`,
`get_random_rotations`, `get_projections_spiral`), each simplified to
`if seed is not None: nx.seed(seed)`.

Two other occurrences use different logic and should be checked
separately, not assumed to be the same bug: `ot/sliced/_spherical_sliced.py:435`
(`not isinstance(...)`, inverted condition) and `ot/utils.py:702`.

### Environment
Not environment-specific -- a pure Python/NumPy logic bug, reproducible
on any platform/version with a recent `ot.backend`.

### Additional context
Found while reviewing PR #838. Will open the fix PR referencing this
issue.

CunjieLee:
## Types of changes

- [x] Bug fix
- [x] Tests
- [x] Code cleanup
- [ ] New feature
- [ ] Documentation update

Closes #848.

`NumpyBackend.seed()` currently forwards every non-`None` value to
`RandomState.seed()`. Passing an existing `np.random.RandomState`
instance therefore raises:

```text
TypeError: Cannot cast scalar from dtype('O') to dtype('int64')
according to the rule 'safe'
```

This forces callers that support both integer seeds and existing
`RandomState` instances to special-case the NumPy backend.

This change makes `NumpyBackend` adopt an externally supplied
`RandomState`, matching the generator-handling behavior of the other
backends. The NumPy-specific `RandomState` branches in the sliced
sampling utilities are removed accordingly.

The fourth occurrence mentioned in #848, in `get_projections_spiral`,
belongs to the still-open PR #838 and is not present on the current
`master` branch. That occurrence can use the unified backend path once
this fix is incorporated into #838.

The issue was reproduced on the unmodified `master` branch using:

```python
import numpy as np
from ot.backend import NumpyBackend

nx = NumpyBackend()
rng = np.random.RandomState(42)
nx.seed(rng)
```

The original code raises the reported `TypeError`.

The same code succeeds after the fix. Additional verification confirms
that:

- `NumpyBackend.rng_` is the exact supplied `RandomState` instance;
- subsequent `nx.randn(...)` calls use that generator's sequence;
- integer seeding retains its previous deterministic behavior.

A regression test was added in `test/test_backend.py`.

The following checks pass:

```text
pre-commit run --all-files
pytest --durations=20 -v test/ --doctest-modules
```

The full test run completed with 1456 passed, 96 skipped, and 4 xfailed.
The skipped tests require optional dependencies that are not installed in
the local test environment.

## PR checklist

- [x] The documentation is up-to-date with the changes I made (no public API documentation changes are required).

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,7 @@
 
 #### Closed issues
 
+- Allow `NumpyBackend.seed` to adopt an existing `np.random.RandomState` instance and remove NumPy-specific random sampling paths in sliced utilities (Issue #848)
 - Preserve input dtype and device for expected sliced plans, avoid materializing dense distance matrices for sparse plans, and fix weighted sparse-distance ordering (PR #846, Issue #845)
 - Fix the sign issue in updates of the previous transport plan in `ot.batch.proximal_bregman_log_plan_batch` (Issue #842)
 - Load triton before TensorFlow in `ot.backend` so that building a torch optimizer no longer segfaults the interpreter, and remove the `torch<2.12` pin from the doctest and documentation requirements (PR #839, Issue #816)
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,7 +8,7 @@
 
 #### Closed issues
 
-- Allow `NumpyBackend.seed` to adopt an existing `np.random.RandomState` instance and remove NumPy-specific random sampling paths in sliced utilities (Issue #848)
+- Allow `NumpyBackend.seed` to adopt an existing `np.random.RandomState` instance and remove NumPy-specific random sampling paths in sliced utilities (PR #849, Issue #848)
 - Preserve input dtype and device for expected sliced plans, avoid materializing dense distance matrices for sparse plans, and fix weighted sparse-distance ordering (PR #846, Issue #845)
 - Fix the sign issue in updates of the previous transport plan in `ot.batch.proximal_bregman_log_plan_batch` (Issue #842)
 - Load triton before TensorFlow in `ot.backend` so that building a torch optimizer no longer segfaults the interpreter, and remove the `torch<2.12` pin from the doctest and documentation requirements (PR #839, Issue #816)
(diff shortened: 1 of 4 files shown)


rflamary:
Hello @CunjieLee #838 has been merged could you hav a quick look and make the new method use nx.seed please?

PR #856, opened 2026-09-14
[MRG] Use ot.utils.check_marginal for default marginals across solvers

tomMoral:
## Types of changes

- [x] Refactoring / consistency (non-breaking, no new user-facing feature)

Follow-up to #851, which introduced `ot.utils.check_marginal` (fill a uniform
marginal on the inputs' backend/device when `None`, otherwise validate shape +
dtype/device) and used it in the batch solvers.

The same `if a is None: a = unif(...) / nx.ones(...) / n` idiom is repeated ~25×
across the rest of the library. Routing those through `check_marginal` makes
default-marginal handling consistent and adds early shape/dtype/device
validation for user-provided marginals (previously a mismatch surfaced as a deep
error further down the solver).

## Description

Replaced the default-marginal idiom with `check_marginal` at 25 backend-preserving
1D sites:

- `ot/gromov/`: `_gw.py`, `_partial.py`, `_semirelaxed.py`, `_lowrank.py`
- `ot/lowrank.py`, `ot/factored.py`, `ot/stochastic.py`
- `ot/bregman/_geomloss.py`, `ot/bregman/_barycenter.py`
- `ot/lp/_barycenter_solvers.py`, `ot/utils.py` (`split_sample_ratio`)

`check_marginal` stays in `ot.utils` (not exported at top-level).

Deliberately **not** changed:
- `ot/gaussian.py` — weights are `(n, 1)` column vectors; `unif`/`check_marginal`
  normalize over the last axis, which would be wrong for that shape.
- `ot/gromov/_dictionary.py`, `ot/weak.py`, `ot/bregman/_empirical.py` — numpy-by-design
  paths (they convert everything to numpy first), so a CPU default is intentional.

One small hardening to `check_marginal`: when validating a provided marginal it
now infers the backend with `get_backend(type_as, a)`, so a cross-backend
marginal raises the standard "Backends should be the same" error instead of an
`AttributeError`.

## Note on behavior

These sites already passed `type_as`, so this is a **consistency + validation**
change, not a device bugfix. The one behavior change is that a provided marginal
whose shape/dtype/device disagrees with the reference now raises early. For the
common `a=None` path and for consistent-backend inputs (all existing tests) the
behavior is unchanged.

## PR checklist

- [x] I have read the [**CONTRIBUTING**](https://github.com/PythonOT/POT/blob/master/.github/CONTRIBUTING.md) document.
- [x] The documentation is up-to-date with the changes I made (check RELEASES.md).
- [x] All tests passed, and additional code has been **covered with new tests** (helper already covered in #851).
- [x] I have added the PR and Issue fix to the RELEASES.md file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rflamary (approved):
Thanks very much @tomMoral

Issue #806, opened 2026-04-23
Faster implementation of Wasserstein Discriminant Analysis


dherrera1911:
## 🚀 Feature
The current WDA implementation is very slow. It is implemented using `numpy` and `pymanopt`. An implementation in PyTorch can be faster.

### Motivation
I was trying to use WDA as a comparison method in my work, but it was almost unfeasible to do so with the POT implementation, and there does not seem to be another good implementation available in Python.

### Alternatives
For my research, I ended up implementing WDA in PyTorch, using POT for the optimal transport objective, and using the POT implementation of wda as a template. My implementation is a lot faster than the current one in POT. My guess is that this is largely because of PyTorch's [native orthogonal constraint](https://docs.pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.orthogonal.html), which uses a  computationally efficient approach of [parametrizations](https://docs.pytorch.org/tutorials/intermediate/parametrizations.html) to impose orthogonality, and allows to use PyTorch's efficient optimizers.

I am wondering if a contribution with this fast WDA implementation would be of interest to POT. It would only work with the PyTorch backend.

rflamary:
yes we are interested. pymanopt is far from optimal so a new pytorch only impelmentation is OK

deeb01:
@dherrera1911 are you still planning to open a PR for this? If not I'd be happy to pick it up.

I've been reading `ot/dr.py` and would want to benchmark before rewriting, to separate the autograd tape overhead from the O(C²) Sinkhorn solves and from pymanopt's retraction.

deeb01:
Benchmarked before proposing. CPU time, n=600, d=10, C=3, seeded harness.

- pymanopt accounts for 0.6–1.8% of runtime, flat from d=10 to d=800, so replacing the retraction recovers ~1%.
- torch beats autograd by 2.5× at best per gradient call, and 0.98× at n=2400. Both sit at ~9× a numpy forward, so the time goes to matmul FLOPs.
- The cost is backprop through the k unrolled Sinkhorn iterations. Freezing the plan (compute `G` outside the tape, differentiate only `<G, M(P)>`) gives **4.3× at k=10 and 11.7× at k=50 in the existing autograd code**.

The frozen-plan gradient is biased: the envelope theorem covers `<G,M> + reg·Σ G log G`, while `wda` minimises `<G,M>` alone. Scored on the true objective under identical Riemannian descent, 3 seeds:

| reg | final objective | grad time | subspace angle |
[...]

PR #858, opened 2026-09-15
[MRG] Add a PyTorch solver to ot.dr.wda (#806)

deeb01:
## Types of changes

New feature (non-breaking change which adds functionality).

Addresses #806, following @rflamary's request for `solver='torch'` alongside the default `solver='autograd'`.

`solver='torch'` runs PyTorch autodiff with Riemannian gradient descent on Stiefel — QR retraction, backtracking with an adaptive initial step — mirroring pymanopt's `SteepestDescent` so both solvers target the same optimum. It needs only torch, and accepts torch tensors directly, keeping device and dtype.

`ot.dr` dependencies are now imported optionally, each function raising an `ImportError` naming what it needs. **This changes `import ot.dr` from raising to succeeding on a partial install** — flagging it in case you prefer the old behaviour.

On speed: pymanopt is 0.6–1.8% of runtime and this solver is 1.0–1.75× faster, so the dependency choice is the real benefit. A frozen-plan gradient for a larger win stalls at a worse objective; numbers on the issue.

Also raises a clear `ValueError` when the between-class cost underflows to zero, previously a divide-by-zero.

New tests assert the torch objective and gradient match the autograd ones (rtol 1e-10, 1e-8) and that both solvers reach a comparable objective from the same start; ratio 0.997–1.004 across n, k, C and `sinkhorn_method`. Also covered: tensor input dtype, `sinkhorn_log`, input not mutated, error path.

`test_dr.py` 16 passed; full suite 2668 passed, 62 skipped, 6 xfailed; pre-commit clean.

## PR checklist


ot/dr.py
```diff
@@ -79,6 +111,184 @@ def split_classes(X, y):
...
+    ui = torch.zeros(M.shape[0], dtype=M.dtype, device=M.device)
+    vi = torch.zeros(M.shape[1], dtype=M.dtype, device=M.device)
+    log_w1, log_w2 = torch.log(w1), torch.log(w2)
+    for _ in range(k):
+        vi = log_w2 - torch.logsumexp(Mr + ui[:, None], 0)
+        ui = log_w1 - torch.logsumexp(Mr + vi[None, :], 1)
+    return torch.exp(ui[:, None] + Mr + vi[None, :])
+
+
+def _stiefel_retract(P, X):
```
rflamary:
```suggestion
def _stiefel_retract_torch(P, X):
```

